PR #82, opened 2026-05-27
feat(ui): ComponentKind.JOB — picker option + JOB badge on deployed releases

taleodor-claude:
## Summary

Pairs with the rearm-saas backend change that adds \`JOB\` to \`ComponentKind\` and treats job-kind components like \`TRANSIENT\`-status deps in product-release matching.

- Component edit page: kind dropdown gets a third option, **Job / CronJob**, next to Generic and Helm.
- Deployed Component Releases (instance page): when the deployed release's component is JOB-kind, render a \`JOB\` badge next to the component name with a tooltip explaining the workload is schedule-driven and the deployed version may legitimately lag the target until the job next runs. Makes it visible *why* the product still shows as matched even with a stale-looking job version.
- \`componentDetails\` GraphQL fragment now includes \`kind\` so the row knows what to render.

## Deferred

Pending-update badge (showing the actual vs target version diff on JOB rows) — requires walking the target product release's \`parentReleases\` which isn't on the current Instance fragment. Will follow if needed.

## Test plan

- [ ] Edit a component; set kind to Job / CronJob; reload — kind persists
- [ ] Open an instance with a JOB-kind deployed component → table shows the JOB badge with hover tooltip
- [ ] Non-JOB rows render unchanged

### DIFF
--- a/ui/src/components/ComponentView.vue
+++ b/ui/src/components/ComponentView.vue
@@ -311,7 +311,7 @@
                                         </div>
                                         <div class="versionSchemaBlock" v-if="updatedComponent && componentData && (componentData.type === 'COMPONENT') && myUser.installationType === 'SAAS'">
                                             <label  id="componentKindLabel" for="componentKind">Component Kind</label>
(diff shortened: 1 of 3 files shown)


taleodor-claude:
Closing — pairs with rearm-saas #135 close, see follow-up PR for the StatusEnum.JOB redesign.

PR #117, opened 2026-06-03
feat(rebom): capture cpe and licenses in component extraction

taleodor-claude:
Part of the **synthetic Dependency-Track registry** feature (rebom half). Pairs with relizaio/rearm-saas (the rearm-core side); **this must deploy first** — rearm-core's `parseBomById` query starts requesting `cpe`/`licenses`, which only exist after this.

## What
- `ParsedBomComponent` gains `cpe` + a normalized/deduped `licenses[]` (SPDX id → name → expression).
- `component.cpe` carried through; `BomComponent` GraphQL type exposes both.
- Purely additive — still only purl-bearing components are emitted.

## Validation
End-to-end validated on psclaude: forced re-parse populated 200 matchable components with licenses (e.g. `["MIT"]`) and cpe carried where present.

## Test plan
- [x] `npm run build` + vitest extractor suite (19/19)
- [x] Live on psclaude via the feature set

### DIFF
--- a/rebom-backend/src/schema.graphql.ts
+++ b/rebom-backend/src/schema.graphql.ts
@@ -147,7 +147,7 @@ const typeDefs = gql`
     version: String
     isRoot: Boolean!
     cpe: String
-    licenses: [String!]!
+    licenses: [Object!]!
   }
 
   type BomDependency {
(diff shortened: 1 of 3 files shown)


taleodor-claude:
Superseded by #120, which carries these cpe/licenses commits plus the required-qualifier preservation (julia/swid/oci) and CPE-canonical component emission needed by the synthetic Dependency-Track work. Closing in favor of #120.

PR #133, opened 2026-06-11
feat(ui): actionable approval events (notifications Phase 4b)

logicflakes:
## Summary

UI half of the actionable approval events feature (Phase 4 of the notifications roadmap; backend landed separately):

- **ReleaseView · Approvals tab**: a write-gated **Request Approvals** button calls the `requestReleaseApprovals` mutation with no entry filter — the backend targets every still-actionable approval entry and errors if nothing is pending (surfaced as-is). Open (unresolved) requests render above the approval matrix with requester, timestamp and entry names.
- **Pro-only schema isolation**: `approvalRequests` / `requestReleaseApprovals` are absent from the OSS schema, so the request list rides a dedicated query gated on `installationType !== 'OSS'` (same pattern as the notification inbox queries) instead of the shared release fragments — folding it into the fragments would break every release fetch on OSS. Fire-and-forget from `fetchRelease` with a stale-response guard for prev/next navigation.
- **NotificationInbox · detail drawer**: typed summary for `APPROVAL_REQUESTED` / `APPROVAL_RESOLVED` payloads — release deep link (closes both drawers on navigation), approval entry names, requester/resolver identity, and an APPROVED/DISAPPROVED tag. The raw-payload collapse remains the generic fallback for other event types.
- **Filter dropdowns**: `eventTypeOptions` gains the two approval event types, plus the three release event types (`RELEASE_CREATED` / `RELEASE_LIFECYCLE_CHANGED` / `RELEASE_BOM_DIFF`) that have been missing from the subscription and inbox filters since they shipped.

## Notes

- `npm run lint` is broken on main independently of this PR (deps bump moved to eslint 10; the script still passes the removed `--ignore-path` flag and the config is legacy `.eslintrc.js`) — left for a separate chore.

## Test plan

- [x] `npm run build` green
- [ ] Sandbox smoke: request approvals on a release with a pending policy → targeted inbox rows appear for eligible approvers; deep link lands on the release; casting the resolving vote emits APPROVAL_RESOLVED and marks the request resolved
- [ ] Verify OSS-safety: release page loads with no approval-requests query when installationType is OSS

### DIFF
--- a/ui/src/components/NotificationHistory.vue
+++ b/ui/src/components/NotificationHistory.vue
@@ -238,14 +238,20 @@ const deliveryColumns = computed(() => [
     },
     {
         title: 'Channel', key: 'channelUuid',
-        render: (row: DeliveryRow) => channelNameById.value[row.channelUuid]
-            || h('span', { class: 'muted-12', title: row.channelUuid }, '(deleted channel)'),
+        // Null channel = Phase 4a targeted delivery (personal inbox copy,
+        // no transmission channel) — not a deleted channel.
+        render: (row: DeliveryRow) => (row.channelUuid
+            ? (channelNameById.value[row.channelUuid]
+                || h('span', { class: 'muted-12', title: row.channelUuid }, '(deleted channel)'))
+            : h('span', { class: 'muted-12' }, 'Direct')),
     },
     {
         title: 'Subscription', key: 'subscriptionUuid',
         render: (row: DeliveryRow) => {
             if (!row.subscriptionUuid) {
-                return h('span', { class: 'muted-12' }, '(channel test)')
+                // Channel tests have a channel but no subscription; targeted
+                // approval deliveries have neither.
+                return h('span', { class: 'muted-12' }, row.channelUuid ? '(channel test)' : 'Direct')
             }
             return subscriptionNameById.value[row.subscriptionUuid]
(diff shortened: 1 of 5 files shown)


logicflakes:
## Sandbox smoke results (Phase 4b)

Deployed this branch's UI to the sandbox and ran the full request → inbox → resolve loop with two users (requester + a separate approver):

- **Request Approvals** on ReleaseView → success toast and the "Open Approval Requests" block renders with entries + requested-by.
- Approver (not the requester — requester correctly excluded) got unread badge `1` and a targeted `APPROVAL_REQUESTED` inbox row; the detail drawer shows the release deep link, approval entries, and requester.
- Approving as that user produced the subscription-routed `APPROVAL_RESOLVED` row with the **Resolution: APPROVED** tag and "Resolved by …" line; the live Teams dispatch on the route went out (SENT, 1 attempt).
- The open-request block on ReleaseView cleared after resolution.

Two bugs surfaced during the smoke, both fixed:

1. **UI labels for targeted deliveries** (this PR, commit ab48bc2f): per-user targeted rows have no channel/subscription, and the inbox + history surfaces mislabeled them as "(deleted)" / "(channel test)". They now render as **Direct**. Re-verified on the sandbox after the fix rolled — drawer shows `Channel: Direct`.
2. **Resolve-marks-read write could be silently lost** — transaction-propagation issue in the backend's after-commit hook; fixed in a separate backend PR.

Smoke verdict: ✅ all flows pass.

PR #136, opened 2026-06-12
feat(ui): CISA KEV badges + KEV details modal (Phase 6b)

logicflakes:
## Summary

- **KEV badge in the findings table**: vulnerabilities listed in the CISA Known Exploited Vulnerabilities catalog get a clickable red `KEV` tag next to their Issue ID in `VulnerabilityModal` — covering ReleaseView (release- and artifact-level), BranchView, MostRecentReleasesWidget, and MarketingReleaseView.
- **New `KevDetailsModal`**: clicking the badge shows the full KEV catalog entry — vendor/project, product, date added, FCEB remediation due date, ransomware-campaign use, CWEs (linked to MITRE), description, required action, notes — plus consent-gated external links to the CISA catalog and osv.dev.
- **New `kevService`**: Pro-only read surface. The `knownExploited` flag and `kevRecordDetails` query are absent from the OSS schema, so (like `approvalRequests`) they live in separate queries gated on `installationType !== 'OSS'` and are never folded into the shared release/metrics fragments. All KEV fetches fail soft (empty set / null) so a findings view never breaks over the badge.

Annotation happens once in the `ReleaseVulnerabilityService.fetchReleaseVulnerabilityData` chokepoint; callers just pass their `installationType`. Changelog finding lists and the analytics/history surfaces (AppHome, FindingsOverTimeChart, VulnerabilityAnalysis) keep their current rendering — KEV badges there are a follow-up.

Depends on the Pro backend KEV read surface (Phase 6a); on OSS or older backends the UI sends no KEV queries and renders exactly as before.

## Test plan

- [x] `npm run build` green (`npm run lint` is broken at main: eslint v10 vs eslint-8 flags in the script — pre-existing, untouched here)
- [x] Two-layer review (`/review` + convention check) — done; layer-2 MUST-FIX (alias-primary rows) + SHOULD-FIXes fixed in 85f82af4, summary in PR comment
- [x] Sandbox verification (2026-06-12): KEV badges on the Log4Shell pair only, badge → details modal with populated CISA fields, clean GraphQL trace — evidence in PR comment (OSS path covered by review; sandbox is a Pro install)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #137, opened 2026-06-12
feat(ui): KEV badges on analysis, changelog, home and trend surfaces

logicflakes:
## Summary

Follow-up to #136: extends CISA KEV visibility to the four remaining UI surfaces:

- **Vulnerability Analysis** (`VulnerabilityAnalysis.vue`) — KEV tag next to the finding id, click opens the KEV details modal.
- **Component & Organization changelogs** — KEV tag on vulnerability findings in all changelog sections (new / partially resolved / inherited debt / still present / resolved), both NONE and AGGREGATED modes.
- **AppHome** most-vulnerable-components widget — vulnerabilities in the drill-in modal carry the KEV badge.
- **Findings-over-time chart** drilldown — same.

## Design: conditional field inclusion instead of a mirror query

#136 used a small mirror query (`fetchReleaseKevVulnIds`) to keep OSS schemas untouched. That pattern is wrong for these surfaces: they sit on expensive backend computations (changelog diffing, `findingsPerDay`, `mostVulnerableComponentsPerOrg`), and a mirror query would re-run each computation in full.

Instead, a new `kevFieldSelection(installationType)` helper returns `'knownExploited'` on Pro and `''` on OSS (or while installationType is unknown), so the flag rides the main query at zero extra cost and OSS queries never reference a field their schema lacks. This is safe because `installationType` comes from the same backend that serves the schema. Both sanctioned patterns are now documented in `kevService.ts`.

Notes:

- `changelogQueries.ts` fragments became functions of the kev field; the file stays store-free — callers pass `installationType` in.
- The vuln-analysis DTO has no `knownExploited`, so analysis rows are stamped by matching `findingId`/aliases against KEV ids collected from the scope's `vulnerabilityDetails` (Pro-only by construction — the set is empty on OSS).
- Rows are annotated via copies, not mutation (Apollo result freezing).

## Test plan

- [x] `npm run build` passes (lint is broken at main, pre-existing)
- [ ] Sandbox verify: KEV tag + modal on Vulnerability Analysis, component changelog, org changelog, AppHome widget, findings-over-time drilldown
- [ ] Confirm no `knownExploited` field errors on OSS (field omitted when installationType ≠ Pro)

### DIFF
--- a/ui/src/components/AppHome.vue
+++ b/ui/src/components/AppHome.vue
@@ -567,6 +567,7 @@ import ComponentBranchesTable from './ComponentBranchesTable.vue'
 import VulnerabilityModal from './VulnerabilityModal.vue'
 import MostRecentReleasesWidget from './MostRecentReleasesWidget.vue'
 import { processMetricsData } from '@/utils/metrics'
+import { kevFieldSelection } from '@/utils/kevService'
 
 const store = useStore()
 const router = useRouter()
@@ -1264,6 +1265,7 @@ async function fetchMostVulnerableComponents () {
                                 analysisState
                                 analysisDate
                                 attributedAt
+                                ${kevFieldSelection(installationType.value)}
                                 aliases { type aliasId }
                                 sources {
                                     artifact
--- a/ui/src/components/ChangelogView.vue
+++ b/ui/src/components/ChangelogView.vue
@@ -250,7 +250,7 @@
                             :branch-uuid="entry.branch.branchUuid"
                             :branch-name="entry.branch.branchName"
                         />
-                        <FindingChangesDisplay :finding-changes="entry.release.findingChanges" />
+                        <FindingChangesDisplay :finding-changes="entry.release.findingChanges" :org-uuid="changelog.orgUuid" />
                     </div>
                 </div>
 
@@ -274,7 +274,7 @@
                                     :branch-uuid="branch.branchUuid"
                                     :branch-name="branch.branchName"
                                 />
-                                <FindingChangesDisplay :finding-changes="release.findingChanges" />
+                                <FindingChangesDisplay :finding-changes="release.findingChanges" :org-uuid="changelog.orgUuid" />
                             </div>
                         </template>
                     </div>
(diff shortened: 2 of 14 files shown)


logicflakes:
## Two-layer self-review — clean

**Layer 1 (/review):** Approve, no blocking issues. Verified: all `kevFieldSelection` interpolation sites fail safe to `''` on OSS/unknown installationType; all 8 changelogQueries fragment→function conversions consistent with no stale references; VulnerabilityAnalysis KEV stamping is copy-based (Apollo freeze-safe), gated to VULNERABILITY rows, empty-set short-circuit on OSS; alias-shape tolerance (plain strings vs `{aliasId}`) confirmed against `resolveKevCveId`.

**Layer 2 (skeptical senior-engineer pass):** No blockers, no should-fixes. Key challenges held up:
- **installationType timing**: no race — `AppWrapper` awaits `fetchMyUser` inside `<Suspense>` before any routed component mounts, and every call site reads the value at fetch time. Apollo `cache-first` in FindingsOverTimeChart is safe because Pro/OSS variants are distinct query documents.
- **Analysis stamping**: suppressed/existing records are stamped too (markKev runs on the combined list); false positive requires sharing an alias, i.e. being the same vuln.

Noted nits (not fixed here):
- Inherited from #136: a KEV-tagged row whose id is GHSA-only with no CVE alias would open the modal in its "not listed" state — pre-existing behavior, candidate follow-up.
- One `KevDetailsModal` per release entry in changelog NONE mode (NModal mounts lazily, negligible).
- Function-local `FINDINGS_FIELDS`/`FINDINGS_DETAILS_FRAGMENT` keep SCREAMING_CASE to minimize the diff.

Next: sandbox verification per the test plan.

logicflakes:
**Sandbox verification — PASS on all follow-up surfaces** (UI image `2026-06-kev-ui-followups.0`, Pro sandbox)

Fixture: fresh component with an ASSEMBLED release + log4j-core 2.14.1 CycloneDX BOM → 7 findings, 2 of them KEV (CVE-2021-44228, CVE-2021-45046).

Playwright probe results:
- **Vulnerability Analysis** (`/vulnerabilityAnalysis/:org`): 2 KEV tags in the Finding ID column; clicking opens the CISA details modal (vendor/project, dates, required action rendered — no load failure).
- **Findings Over Time → Organization Changelog modal** (`/findingsOverTime/:org`): 2 KEV tags in the Finding Changes tab (aggregated view); KEV tag click opens the same CISA modal from inside the changelog.
- **AppHome most-vulnerable widget**: severity-circle click opens the findings modal with 2 KEV tags.
- **Zero GraphQL errors** across all pages — confirms the conditional `knownExploited` field inclusion produces valid queries on a Pro installation. (OSS leg already covered by the `kevFieldSelection('')` unit of behavior: the field name is simply absent from the selection set.)

Screenshots captured for all three surfaces + both modals.

PR #138, opened 2026-06-12
Clarify Sentinel channel label: Microsoft Sentinel (Azure Log Analytics)

logicflakes:
## Summary
- Resolves the deferred Sentinel-channel naming decision: keep the `SENTINEL` enum value (zero migration), clarify the user-facing labels instead.
- Channel-type dropdown option: "Azure Sentinel" → "Microsoft Sentinel (Azure Log Analytics)".
- Type chip label: "Sentinel" → "Microsoft Sentinel".

The channel ships events to an Azure Log Analytics workspace; Sentinel is the SIEM layered on top, but any Log Analytics workspace works. The label now says both, while keeping the brand name security teams look for.

## Test plan
- [x] `npm run build` green (label-only change in `notificationsCommon.ts`)

### DIFF
--- a/ui/src/components/OrgIntegrations.vue
+++ b/ui/src/components/OrgIntegrations.vue
@@ -236,7 +236,7 @@
                 <div>
                     <div class="info-banner-title">Notification subscriptions</div>
                     <div class="info-banner-body">
-                        Rules that route security and operational events to your messaging channels. Channels themselves are configured as Slack / Teams / Webhook / Sentinel integrations in the Catalog.
+                        Rules that route security and operational events to your messaging channels. Channels themselves are configured as Slack / Teams / Webhook / Microsoft Sentinel integrations in the Catalog.
                     </div>
                 </div>
             </div>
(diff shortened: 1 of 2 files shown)


logicflakes:
Self-review (`/review`) notes:

- **`TYPE_LABELS` change is the live surface** — channel chips in NotificationHistory.vue, SubscriptionsOfOrg.vue, and ChannelGroupsOfOrg.vue all render `name (TYPE_LABELS[type])`, so Sentinel channels now read "(Microsoft Sentinel)". Raw-enum fallback preserved.
- **`typeOptions` is currently a dead export** — nothing imports it since the standalone channel-create dialog was dropped in favor of the Integrations Catalog, and the Catalog card list has no SENTINEL card yet. Kept anyway: it records the canonical dropdown label ("Microsoft Sentinel (Azure Log Analytics)") for when a Sentinel/Webhook create surface lands. Same applies to `webhookAuthOptions`.
- **Fixed in 649f5135**: the subscriptions info banner in OrgIntegrations.vue was the one other user-facing string still saying bare "Sentinel" — now "Microsoft Sentinel".

String-only change, no logic/schema/migration impact. `npm run build` green (repo-wide `npm run lint` is broken pre-existing — eslint `--ignore-path` mismatch).

PR #142, opened 2026-06-14
feat(ui): KEV detail modal to per-source assertion contract

logicflakes:
## Summary

Move the KEV detail surface to the new per-source assertion contract from the backend.

`kevRecordDetails` now returns multiple per-source assertions plus an aggregated ransomware status (`KNOWN`/`UNKNOWN`/`UNSPECIFIED`) instead of one flat CISA record. This updates the service types + query and the detail modal:

- Descriptive fields render from the first active assertion.
- Ransomware shown from the aggregated enum (with optional campaign links) rather than a Known/Unknown boolean.
- New **Reported by** row lists each source; a withdrawn source shows a `revoked <date>` note (the CVE still counts as KEV).
- CISA catalog link shown only when CISA is among the sources; copy de-CISA-fied since KEV is no longer single-source.

Paired with the backend KEV redesign (per-source, soft-deletable assertions). KEV gating is unchanged here — this is the read-surface contract update.

## Test plan

- [x] `npm run build` (vue-tsc) green
- [x] Verified on the sandbox against the redesigned backend: detail modal renders the new per-source layout, ransomware enum, and the Reported-by row
- [x] Soft-delete: with a source revoked, the badge persists and the modal shows the `revoked <date>` note (browser smoke, active + revoked states)


logicflakes:
**Two-layer review done; post-review fix pushed.**

No blockers. Nit fixed in the follow-up commit: ransomware labels disambiguated — `UNKNOWN` → "No known use", `UNSPECIFIED` → "Not reported" (previously "Not known"/"Unknown" read as a typo). `npm run build` green.

logicflakes:
Reconciled with main after the notifications CE catch-up merged: merged `main` in cleanly (no conflicts — KEV UI files don't overlap). `npm run build` green.

PR #165, opened 2026-07-02
fix(ui): split hidden 3-state disapprove checkbox into separate controls

logicflakes:
## Summary
Board task: Approvals disapprove hidden 3-state checkbox cycle -- separate control/legend.

The approval matrix used a single checkbox overloading Naive UI's `indeterminate` state to mean "disapproved" (styled red via a CSS-var override). Disapprove was only discoverable by clicking an already-approved checkbox a second time -- nothing on screen indicated the option existed.

Replaced with two adjacent icon buttons (green check / red X), independently clickable, mutually exclusive. Added a small legend above the approval table explaining them. Disabled/admin-override/comment-field behavior is unchanged -- only the vote control itself changed.

## Verification
Built locally and verified live against the claude2 sandbox with Playwright, using a seeded release with a real open approval request (not a mock):
- Both icons render with the legend visible
- Colors switch correctly: green only when approved, red only when disapproved, gray at rest
- Mutual exclusivity confirmed (clicking one clears the other)
- Click-to-clear confirmed (clicking the active icon again resets to unset)
- Save/Reset Approvals controls appear correctly once there's a pending change

Caught and fixed one real bug during verification, not just eyeballing the diff: `NIcon` takes its color via a dedicated `color` prop, not `style.color` -- the component computes and overwrites its own inline style internally, so the first version silently rendered both icons in a default blue instead of green/red. Confirmed via `getComputedStyle` in the same Playwright run before and after the fix.

## Test plan
- [x] Live-verified on sandbox: render, color states, mutual exclusivity, click-to-clear, save/reset visibility
- [x] Two-layer self-review (see PR comments)

### DIFF
--- a/ui/src/components/ReleaseView.vue
+++ b/ui/src/components/ReleaseView.vue
@@ -1265,7 +1265,7 @@ import { Icon } from '@vicons/utils'
 import { BoxArrowUp20Regular, Info20Regular, Copy20Regular, QuestionCircle20Regular, ChevronLeft20Regular, ChevronRight20Regular } from '@vicons/fluent'
 import { UpCircleOutlined } from '@vicons/antd'
 import type { SelectOption } from 'naive-ui'
-import { NBadge, NButton, NCard, NCheckbox, NCheckboxGroup, NDataTable, NDropdown, NForm, NFormItem, NRadioGroup, NRadioButton, NSelect, NSpin, NSpace, NTabPane, NTabs, NTag, NText, NTooltip, NUpload, NIcon, NGrid, NGridItem as NGi, NInputGroup, NInput, NSwitch, NDatePicker, useNotification, useLoadingBar, NotificationType, DataTableColumns, NModal, NDynamicInput } from 'naive-ui'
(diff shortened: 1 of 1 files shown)


logicflakes:

**Layer 1 (correctness):** Read the full render closure, `updateMatrixCheckbox`, disabled/admin-override logic, and `naive-ui`'s `Icon.mjs` source directly (not just the diff). One real finding: the `NCheckbox` import was now dead code (no longer referenced anywhere in the file) -- removed. Everything else checked out clean: no cross-iteration closure aliasing, mutual exclusivity has no invalid-both-set state even under a click race, disabled guards match old semantics exactly, and the `color`-prop fix (vs. `style.color`) was independently confirmed correct by reading how `NIcon` merges its own computed style over any raw `style` attr.

**Layer 2 (ReARM conventions):** One real finding: the two new icons used bare `class: 'clickable'` where every other clickable `NIcon` in this file uses `class: 'icons clickable'` -- fixed to match. ASCII, `data-testid` naming, tooltip conventions, disabled-opacity styling, and the `@vicons/tabler` import (not a new dependency) all checked out clean.

Both layers re-run clean after fixes. Ready for human review.

PR #166, opened 2026-07-02
fix(ui): explain why 'Needs my approval' is empty for an admin with no role

logicflakes:
## Summary
Board task: Approvals "Needs my approval" empty-state hint (approver role vs org-admin).

An Org Admin can vote on any release (unconditional capability), but the "Needs my approval" tab -- and the `APPROVAL_REQUESTED` notification -- is deliberately keyed off an explicit approval role, not the ADMIN override. An admin with no approval role correctly sees an empty list even when releases are pending elsewhere, and the generic "No Data" empty state gave no hint why -- this reads as broken if you don't already know the distinction.

Added a tailored empty-state hint (via the data-table's `#empty` slot) that only shows for an org admin with zero approval-role grants, explaining the split and what to do about it. Anyone else gets a plain "nothing pending" message instead.

## Verification
Verified live against the claude2 sandbox with two real accounts in the same org (not mocked):
- An org-admin account with no approval-role grant: tailored hint shown, generic message not shown, confirmed via screenshot and `data-testid` visibility check.
- An account holding the relevant approval role, with an actual pending approval request open on a real release: the row renders, no hint shown, confirmed the same way.

## Test plan
- [x] Live-verified both empty-state branches with two distinct real accounts and a real pending approval request
- [x] Two-layer self-review (see PR comments)

### DIFF
--- a/ui/src/components/NotificationInbox.vue
+++ b/ui/src/components/NotificationInbox.vue
@@ -124,7 +124,19 @@
                         :single-line="false"
                         :bordered="false"
                         :row-key="(row) => row.releaseUuid"
-                    />
+                    >
+                        <template #empty>
+                            <n-alert
+                                v-if="isOrgAdmin && !hasApprovalRoleGrant"
+                                type="info"
+                                :show-icon="true"
+                                data-testid="approvals-empty-admin-hint"
+                            >
+                                Nothing is waiting on your approval role right now. As an Org Admin you can still approve or disapprove any release directly from its page -- this list only shows releases where you hold an explicit approval role. Ask another admin to grant you one if you want items to appear here automatically.
+                            </n-alert>
+                            <span v-else data-testid="approvals-empty-generic">No releases are waiting on your approval right now.</span>
+                        </template>
+                    </n-data-table>
                 </n-tab-pane>
                 </n-tabs>
             </n-drawer-content>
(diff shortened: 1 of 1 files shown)


logicflakes:

**Layer 1 (correctness):** Verified the `myuser.permissions.permissions[]` shape against `graphqlQueries.ts` and confirmed the frontend/backend gating logic match (`obtainCombinedUserOrgPermissions` is the same method behind both `ApprovalNeedsService` and the `user` query). No flash of the wrong hint during initial load (naive-ui hides the empty slot while `loading=true`, set synchronously). One real bug found: the notification drawer is mounted once at nav level and can stay open across an org switch -- `watch(orgUuid)` reset the approvals list to empty but only refreshed the badge count, not the full row list, so a still-open drawer on a newly-switched org would show the new specific empty-state hint asserting "nothing pending" for data that was simply never fetched. Before this PR that was just a stale generic placeholder; this PR turned it into a specific, potentially false claim. Fixed by also calling `loadApprovals()` in the watcher when the drawer is open on the approvals tab, reusing the existing `approvalsInflightToken` staleness guard so no new race is introduced.

**Layer 2 (ReARM conventions):** Two real findings: (1) `isOrgAdmin` hand-rolled a permission scan duplicating `commonFunctions.isAdmin`, which is already exported and already imported in this file -- reused it instead. (2) The new `n-alert` used `:show-icon="true"`, while every other alert in this file and the codebase's info-type alerts use `:show-icon="false"` -- matched. ASCII and `data-testid` naming checked out clean; this is the first `#empty` slot usage in the codebase so no precedent to diverge from there.

Re-verified live after all fixes (two real accounts, one org-admin-no-role and one role-holder-with-pending-item) -- both empty-state branches still render correctly.

Both layers re-run clean. Ready for human review.

PR #167, opened 2026-07-03
feat(ui): add "Send test notification" button to channel cards

logicflakes:
## Summary
Board task #31: `testNotificationChannel` has been fully implemented server-side since the notifications framework shipped (bypasses subscription matching/dedup, always produces a visible delivery via the outbox -> fan-out -> worker pipeline), but there was no UI to trigger it -- every channel type's config could only be verified by waiting for a real event to route to it.

Adds a "Send test notification" icon to each channel card's instance row, alongside the existing enable/disable, edit, and delete actions. On click: calls `testNotificationChannel(channelUuid)`, then polls `notificationDeliveries(orgUuid, eventUuid, channelUuid)` at a 1.5s cadence for up to 60s for the terminal delivery status, showing a spin animation while in flight. Resolves to a success toast (SENT/ACKED), an error toast with the delivery's `lastError` (FAILED), or a "still pending, check History" toast if the poll window elapses first.

## Related
- rearm-core PR (test-delivery-fast-fail): a broken webhook was taking over an hour to reach FAILED (borrowed backoff curve from an unrelated subsystem) -- fixed so test deliveries fail fast, found while validating this button live.
- rearm-playwright PR (send-test-affordance): flips the harness's B7/B8 tests and adds B9 exercising the real click-to-delivery round trip.

## Verification
- Built locally and live-verified against the claude2 sandbox via the real feature-set CI/deploy pipeline (not a mocked/interception-based test) using two real webhook channels seeded server-side (one deliverable, one broken)
- Confirmed: icon renders on every channel-card row, spins while in flight, resolves to correct success/failure toast, mutual state doesn't leak across rows
- Operator walkthrough completed live on the sandbox, confirmed working (including the paired backend fast-fail fix)

## Test plan
- [x] Live-verified end-to-end on the deployed sandbox (real CI build, not local override) for both success and failure paths
- [ ] Two-layer self-review (in progress)

### DIFF
--- a/ui/src/components/OrgIntegrations.vue
+++ b/ui/src/components/OrgIntegrations.vue
@@ -2150,19 +2150,6 @@ function openEditChannelForCard(card: CardConfig, ch: ChannelCatalogRow) {
 // without needing to reassign a new Set each time.
 const testingChannels: Ref<Set<string>> = ref(new Set())
 
-const TEST_NOTIFICATION_CHANNEL_MUTATION = gql`
-    mutation testNotificationChannel($channelUuid: ID!) {
-        testNotificationChannel(channelUuid: $channelUuid) { uuid status }
-    }
-`
-const TEST_DELIVERY_POLL_QUERY = gql`
-    query testDeliveryPoll($orgUuid: ID!, $eventUuid: ID!, $channelUuid: ID!) {
-        notificationDeliveries(orgUuid: $orgUuid, eventUuid: $eventUuid, channelUuid: $channelUuid, limit: 1) {
-            items { status lastError }
-        }
-    }
-`
-
 // testNotificationChannel bypasses subscription matching/dedup and always
 // produces a visible delivery, but dispatch is still async (fan-out +
 // worker), so the mutation itself only returns a PENDING outbox event uuid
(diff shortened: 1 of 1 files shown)


logicflakes:

**Layer 1 (correctness):** Found and fixed a real bug: `sendChannelTest`'s poll loop read `orguuid.value` (a live computed) fresh on every iteration. This component is reused, not remounted, across an org switch -- switching orgs mid-poll would send the query against the wrong org, silently forcing a "still pending" timeout even if the test had actually resolved. Fixed by snapshotting the org at test-start and abandoning quietly (no toast) if the user switches away mid-poll. Everything else checked out clean: the double-click guard is reliable (synchronous `Set.add` before the first `await`), the poll query can't be shadowed by an unrelated delivery (filtered by both unique `eventUuid` and `channelUuid`), error handling can't leave the icon permanently stuck (try/finally covers the early-return path too), and the icon doesn't reintroduce the `color`-prop-vs-`style` bug from PR #165 (uses CSS class only, no inline color).

**Layer 2 (ReARM conventions):** Two real findings, both fixed: (1) the two new GraphQL operations were hoisted to module-level constants -- the only hoisted `gql` tags in this file, where 30+ other operations are all inlined at their call site. Inlined both, and renamed the poll query's operation name to match the field it calls (`notificationDeliveriesForChannelTest`), like every sibling operation. (2) The spin animation reinvented an identical `@keyframes` that already exists globally as `spin` in `App.vue` -- reused it instead.

Re-verified live after all fixes on the redeployed sandbox: success and failure paths both still resolve correctly, spin animation confirmed using the reused `spin` keyframe (via `getComputedStyle`), no console errors.

Both layers re-run clean. Ready for human review.

PR #186, opened 2026-07-12
fix(logging): SECURITY alert only when both agentic read auth paths deny

taleodor-claude:
Follow-up to a false alarm observed on the controlling ReARM: an agent legitimately reading an org release via the RESOURCE-on-release fallback produced `SECURITY: agentic-session-read denied` at ERROR level, looking like an intrusion while the read was in fact authorized.

- Session-path denials in `AuthorizationService.isFreeformKeyAuthorizedForAgenticSessionRead` now log at INFO with a note that the caller may fall back (matching the already-INFO step-1 org-permission check).
- The SECURITY ERROR now lives in `ReleaseDatafetcher.agenticReleaseProgrammatic` and fires only when both the agent path and the RESOURCE path deny, logging both denial reasons.

No behavior change to authorization itself.


taleodor-claude:
Closing per review — this change lands via rearm-saas #290 only; OSS backend will be reconciled separately.

PR #197, opened 2026-07-17
fix(ui): stop offering PREVIEW subscription status -- it silently does nothing

logicflakes:
## The bug

`PREVIEW` subscription status **does nothing at all** -- it is byte-for-byte identical to `DISABLED`.

`NotificationSubscriptionRepository.findActiveByOrgString` filters on `status = 'ACTIVE'`, so a PREVIEW subscription never reaches fan-out, no delivery row is ever written, and `NotificationDeliveryStatus.PREVIEW` ("render captured for history, never dispatched") has **zero writers**.

This was never built -- `git log -S"PREVIEW"` on `NotificationFanOutService` returns nothing. Not a regression.

## Reproduced on the sandbox

Backend `26.07.40`. One synthetic event fired three times at the same channel, changing **only** the subscription status:

| sub status | outbox | delivery rows |
|---|---|---|
| ACTIVE | FANNED_OUT | **1** |
| PREVIEW | FANNED_OUT | **0** |
| DISABLED | FANNED_OUT | **0** |

A fourth independent ACTIVE control also produced 1 row. Delivery-status census on the instance is `SENT, FAILED` only -- no PREVIEW row has ever existed. Sandbox restored to baseline afterwards (44 deliveries, 0 injected events).

<details>
<summary>Two confounds that would each have faked a PASS</summary>

- **`channel_test_target`**: if non-null, fan-out short-circuits at `NotificationFanOutService:214` into `insertChannelTestDelivery()` and returns, **bypassing the subscription matrix entirely**. The newest event on the sandbox is a channel test -- cloning it naively produces a delivery row in all three trials and "proves" PREVIEW works. Nulled it.
- **Dedup**: fan-out suppresses repeats of the same dedup key within a window, so three identical events would suppress trials 2 and 3 for the wrong reason. Used `origin=SYNTHETIC`, which bypasses dedup (`:643`).

Also verified `findActiveByOrg` is not cached, so the SQL status flip is genuinely visible to fan-out.
</details>

## Why the old label made it worse than a missing feature

`Preview (rows land but no dispatch)` promised a dry run. A user sets PREVIEW, sees nothing land, and concludes **their filter matches nothing** -- then goes and debugs a filter that fan-out never consulted. Two help texts asserted the same falsehood: the History toolbar said PREVIEW rows are included (none can exist), and the test-event dialog warned that other PREVIEW subscriptions would also fire and deliver (they cannot).

## What this PR does

Operator decision: **defer the feature, close the trap.** UI-only; both backend enum values are deliberately kept. This is the same treatment `VEX_STATE_CHANGED` already gets three lines away, for the identical reason ("no event producer in the backend yet, so a subscription on it would never fire (a silent trap)").

- `subscriptionStatusOptions`: PREVIEW becomes `Preview (not yet available)`, `disabled: true`.
- Subscription table: legacy PREVIEW rows tag **`warning`** alongside DISABLED instead of a calm `info` -- the badge shouldn't imply it's doing something benign-but-active.
- Fixed the two false help texts, and added a field hint mirroring the VEX one.
- **Spec guards** pinning PREVIEW and VEX_STATE_CHANGED unselectable.

### Legacy PREVIEW subscriptions are not trapped

The VEX case needed an `eventTypeOptionsForForm` computed because naive-ui sets `closable: !option.disabled` on **tags**, which would strand an existing selection. That doesn't apply here, verified in `node_modules`:

- `Selection.mjs:604` -- `closable: !option.disabled` is inside `if (multiple) {` (`:590`). Status is a **single** select: no tags.
- `Select.mjs:294-303` -- `selectedOptionRef` resolves the label from `valOptMap`, built from **all** options including disabled. A legacy `PREVIEW` value still renders "Preview".
- `SubscriptionsOfOrg.vue:597` -- `toggleSubscriptionStatus` computes `ACTIVE ? DISABLED : ACTIVE`, so a PREVIEW row's Enable button already moves it to ACTIVE.

So a legacy PREVIEW row is rescuable from two directions and can never be re-selected once moved.

### Deliberately left alone

`deliveryStatusOptions`' PREVIEW filter and `deliveryStatusTagType`'s PREVIEW mapping mirror the backend enum on the **read** side. They're unreachable rather than misleading -- a filter returning zero rows is an empty result, not a promise -- consistent with the existing leave-as-is call on the equally writerless `RATE_LIMITED`.

## Verification

- `npm run build` green. `npm test` **40 passed** (was 37).
- The new guards were **mutation-tested**: reverting the fix makes the PREVIEW guard fail; restoring it goes green. A test that can't fail is theatre. This matters because `build` is `vite build` -- **no typecheck** -- so nothing else would catch a silent revert.
- No open PR touches these three files; `#162` (data-testid on notif components) is already merged.

## Review gate

- **Layer 1 `/code-review`** -- 4 findings, all fixed: the `info` badge, the missing regression guard, backend jargon leaking into user-facing hint copy, and a 17-line comment that narrated the diff instead of stating the constraint.
- **Layer 2 conventions subagent** -- independently verified the naive-ui claim on all three legs (it would have been blocking if wrong), confirmed the label/hint/placement conventions match both in-file precedents, and found the `info` badge.
- No `/security-review`: no auth/secrets/input-handling/crypto surface.

## Follow-up

Board task filed capturing the product questions PREVIEW would need if it's ever wanted (window length, auto-promote, rate-limit interaction, inbox visibility, retention), so the spec work isn't rediscovered. `NotificationReadService:67` already hints at unresolved read semantics ("mark all on this subscription as read after PREVIEW window").


logicflakes:
Conflicts resolved and pushed — the branch is mergeable again.

The conflicts came from **#270**, which landed in the same four files. One real conflict, in `notificationsCommon.spec.ts`: both sides appended to the same import list. Resolved as the union.

The other three auto-merged, so I checked by hand rather than trusting it — both PRs' intents are intact:

| File | This PR | #270 |
|---|---|---|
| `notificationsCommon.ts` | PREVIEW disabled, "not yet available" | classifier helpers |
| `SubscriptionsOfOrg.vue` | `warning` status tag, corrected test-dialog help text | reordered poll loop, owner-routing caveat |
| `NotificationHistory.vue` | corrected help text | — |

**Mutation-verified the tripwire still trips post-merge:** re-enabling PREVIEW in `subscriptionStatusOptions` turns *"keeps PREVIEW unselectable while fan-out matches ACTIVE only"* red. That guard is the load-bearing part of this PR and it survived.

135 tests pass, lint clean, build clean.

---

One note for the record, since it nearly caused damage: I independently rediscovered this bug today while doing unrelated cleanup, re-reproduced it on the sandbox, and started building a **worse** fix — removing `PREVIEW` from `NotificationSubscriptionStatus`, `NotificationDeliveryStatus` and both GraphQL enums, plus a migration to rewrite existing rows. That would have broken the API for any client sending `PREVIEW` and needed a migration for nothing.

This PR's approach is better on every count: UI-only, both enum values deliberately kept, and PREVIEW left visible-but-disabled mirroring the `VEX_STATE_CHANGED` treatment three lines away, so a user who had it selected can see *why* it's gone. All of my version has been reverted.

PR #209, opened 2026-07-21
feat(compose): bundled zot OCI registry, on by default, configured via .env

taleodor-claude:
## What

A single `docker compose up` can now include a local OCI registry for ReARM's artifact storage, instead of requiring an external one:

```sh
docker compose -f docker-compose.yml -f docker-compose.zot.yml up -d
```

The overlay adds two services -- a one-shot htpasswd generator and a **single-container zot** (`zot-minimal` v2.1.18, digest-pinned) -- and redirects `oci-artifact-service` at it (`REGISTRY_HOST=zot:5000`, plain http in-network). Design points:

- **Local-only by construction**: zot publishes no host ports; only compose services can reach it.
- **Zero-config**: credentials default to `rearm`/`zotRegistryPass`, overridable via the existing `OCI_REGISTRY_USERNAME`/`OCI_REGISTRY_TOKEN` variables (htpasswd is regenerated from them on every start, so rotation is just an `.env` edit + `up`). Registry namespace defaults to `rearm` for backend and rebom.
- **Tuned to the access pattern** (push once, fetch by digest, no deletes, monthly repo rotation): dedupe off, GC off, no UI/search extensions -- comfortable well past 200k artifacts on a local volume.
- External-registry deployments are untouched -- the overlay only applies when explicitly added.

README gains the scenario plus a note on the inotify startup requirement zot shares with traefik (exits with "couldn't initialize inotify" on inotify-exhausted hosts; fix is `fs.inotify.max_user_instances`).

## Validation

Booted the full stack with the overlay on the sandbox host (latest main images): all nine services healthy, UI 200, `/api/healthCheck` OK. zot rejects unauthenticated `/v2/` (401) and accepts the generated credentials (200). Real round-trip through `oci-artifact-service` using ReARM's exact parameter shapes -- push to `rearm/downloadable-artifacts-2026-07` with a `rearm-*` tag, pull by `tag=sha256:<digest>` (the `SharedArtifactService` retrieval form) -- returns byte-identical content, and the stored manifest is a proper OCI 1.1 artifact (`artifactType` preserved). Merged-config check confirms env overrides apply per-variable and no ports are published.


taleodor-claude:
Reworked per review discussion: the overlay file is gone -- zot is now part of the main compose file, **on by default**, with all configuration via .env. External registries: set the OCI_* variables plus `OCI_USE_PLAIN_HTTP=false`, and optionally `OCI_BUNDLED_REGISTRY_REPLICAS=0` to not run the bundled registry (compose has no env-driven service conditional; a replicas toggle is the disable mechanism since profiles can only opt in, not default on). One upgrade note for existing external-registry deployments: `USE_PLAIN_HTTP` now defaults to true, so they need the one-line `OCI_USE_PLAIN_HTTP=false` addition. Re-validated both merged-config modes and a live push/pull round-trip in default mode.

taleodor-claude:
Ran the reported scenario end to end on a fresh default stack (kcadm user -> unseal -> org -> component -> release -> uploads) and found **two real bugs**, both now fixed on this branch:

1. **The reported "connection refused" is a pre-existing compose bug**: `RELIZAPROP_OCIARTIFACTS_SERVICE_HOST` is the backend's WebClient base URL verbatim (`relizaprops.ociArtifacts.serviceUrl`), and the compose file passed a bare `oci-artifact-service` -- so the backend dialed **port 80** instead of 8083. Any external-registry compose deployment hit this too; the zot default just made it visible. Now `http://oci-artifact-service:8083`.

2. **`oci-artifact-service` used the detected MIME type verbatim as the OCI layer `mediaType`** -- for files detected as e.g. `text/plain; charset=utf-8` that violates the OCI descriptor pattern. Harbor/distribution tolerate it; zot correctly rejects the manifest. Parameters are now stripped at detection (the `artifactType` path already did this).

Verified after the fixes: CycloneDX SBOM upload lands in zot via rebom (`rearm/rebom-artifacts-2026-07`), a text document lands via the backend direct path (`rearm/downloadable-artifacts-2026-07`, clean `text/plain` on both artifactType and layer), pull-by-digest round-trips byte-identical.

**One follow-up needed after merge**: the compose still pins `oci-artifact-service:26.03.12`, which carries bug 2 -- JSON artifacts (SBOMs, the common case) work with the pinned image, but text-like uploads need the pin bumped once CI releases the fixed service.

Also noted while reproducing your boot: a failed first `up` (e.g. host port conflict) can leave `rearm-postgresql` running with **no network attached**, and a plain re-`up` does not heal it (backend then crash-loops on `UnknownHostException: rearm-postgresql`; the UI additionally caches the dead backend IP -> 502 until restarted). `docker compose up -d --force-recreate rearm-postgresql` fixes it. If you saw connection-refused only, you did not hit this one -- but it may be worth a troubleshooting note in the README; say the word and I will add it.

PR #217, opened 2026-07-25
landing: add JFrog AppTrust comparison; broaden Dependency-Track to v4 and v5

taleodor-claude:
Adds a **ReARM vs JFrog AppTrust** section to the comparisons page, placed after Chainloop (its closest peer — both are release-governance rather than scanner comparisons).

## Framing

AppTrust's model — applications, application versions, and versions composed of other application versions — converges on the Product-Component pattern ReARM has run for years. The section says so explicitly and treats that as validation of the approach rather than making any origin claim. The wedge is placed where it is durable and verifiable: **what can be a node** in the hierarchy, and what the platform understands about the levels.

## The six points

| Dimension | ReARM | AppTrust |
|---|---|---|
| Hierarchy | Typed: Products, Components, Feature Sets, Branches, Releases — behavior attaches to a level | Generic recursive container; nesting is a convention you impose |
| Findings | Own audit engine, scoped org → product → feature set → component → branch → release, any source | Lifecycle policy waivers on promotion — passing a gate, not a scoped disposition |
| Parallel maintenance | Branches + Feature Sets: concurrent shipping versions, each with its own patch stream | Lifecycle stages are a progression axis; concurrent streams have no first-class model |
| What can be a node | Any release, including supplier/OEM software known only from a received SBOM (CRA, FDA 524B) | Artifactory-bound: children are builds/artifacts in Artifactory |
| Assembly | Auto-integration via feature sets and dependency patterns | Package bindings solve artifact→application; product assembly stays manual |
| Coupling | Registry- and CI-agnostic, FOSS Community Edition | Anchored to the JFrog platform |

## Accuracy notes

- **No pricing** on either side, per request.
- Claims are written to hold **regardless of two things not confirmable from public sources**: whether AppTrust's nesting is arbitrarily recursive or single-level, and whether aggregate SBOM roll-up traverses the full tree. Nothing asserts a depth limit or a roll-up gap.
- The two rows most exposed to correction if JFrog's surface has moved are *policy waivers* and *manual assembly*; both are worth re-verifying on a trial instance before being used in a talk.

Table of contents is generated from the same array, so the new anchor appears automatically. Page meta description now lists AppTrust. File stays ASCII-only, matching the rest of the page. Astro build passes (46 pages).


taleodor-claude:
**Second commit** broadens the Dependency-Track section from v4-only to both generations (`ReARM vs Dependency-Track`, right label `Dependency-Track (v4 and v5)`). The section id is unchanged, so existing anchors and inbound links keep working.

Every claim was re-verified against a live **Dependency-Track 5.0.2** instance — all 151 API paths, the deployed chart values, the shipped analyzer set, and the running container's file storage — rather than carrying v4-era copy forward.

**Verified, kept:** CycloneDX-only ingestion (zero SPDX paths in the v5 API); uploads not preserved (file storage empty after upload; the export endpoint regenerates a BOM from current project state rather than returning what was submitted); no SARIF/signature/attestation storage; no release approval workflows; enrichment not stored alongside the original.

**Updated where v5 genuinely moved:**
- *Collection projects* (`AGGREGATE_DIRECT_CHILDREN`, `AGGREGATE_DIRECT_CHILDREN_WITH_TAG`, `AGGREGATE_LATEST_VERSION_CHILDREN`) now roll child metrics up the tree. The old "limited single-parent hierarchy" line is replaced by one that credits this while keeping the two real limits: a project has a single parent, and what aggregates is metrics rather than an assembled, versioned product release.
- The audit row is now precise — a `(project, component, vulnerability)` triple or its policy-violation equivalent, across findings from the **internal, OSS Index, Snyk and Trivy** analyzers v5 ships, with CEL vulnerability policy bundles for automation. This replaces wording that wrongly implied other scanners were out of scope; Dependency-Track does integrate with them.

**Dropped:** "significantly larger infrastructure footprint required" — no longer defensible now that v5 ships a `monolith` topology (one Deployment plus Postgres). The per-project SBOM duplication argument replaces it and is the stronger point.

Left-hand (ReARM) cells are unchanged throughout. Astro build passes, file stays ASCII-only, and the AppTrust section from the first commit is unaffected.

PR #236, opened 2026-07-28
fix(compose): let the Keycloak admin credentials be set, and say so

taleodor-claude:
Follows #235 (which is still open — this branches off main, so the two touch the same file and the second to merge will need a trivial rebase).

## The thing worth fixing

The docs told readers they could change the Keycloak admin account. On compose they could not:

```yaml
- KEYCLOAK_ADMIN=admin
- KEYCLOAK_ADMIN_PASSWORD=admin
```

Hardcoded in `docker-compose.yml`, absent from `.env.example`. The only way to change them before first boot was editing a tracked file. And the admin console sits at `<host>/kauth`, reachable from wherever the deployment is — so a non-localhost stack shipped with a well-known password to the identity provider that guards everything else.

Both now read from `KEYCLOAK_ADMIN_USER` / `KEYCLOAK_ADMIN_PASSWORD`, **defaulting to `admin`** so nothing changes for an existing deployment or a localhost evaluation, and documented in `.env.example`.

Like the realm import, these are bootstrapped on **first start only** — a detail worth stating, since setting them later looks like it should work and silently does not. Both files now say so, with the console fallback for an already-running stack.

Verified with `docker compose config`: defaults preserved with no `.env`, and overrides honoured when set.

## Docs

The warning goes where the credentials are first mentioned, rather than in a security section nobody reads at install time. It carries the setting for each path (`.env` for compose, `keycloak.secrets.adminpassword` for Helm — the chart already exposed it) and notes the reset happens in the **master** realm, because the very next step switches to Reliza and it would be easy to look in the wrong place.

The remote-host compose example also gains a `KEYCLOAK_ADMIN_PASSWORD=...` line, since that is the point where someone is deciding what to put in `.env`.

## Helm carve-out dropped

The login URI step said it was required "whenever your host differs from the chart default". The default is a placeholder, so in practice it is required every time — it now says that, and the section's "Applies to" line matches.

### DIFF
--- a/deploy/docker-compose/.env.example
+++ b/deploy/docker-compose/.env.example
@@ -59,6 +59,14 @@
 #OCI_REGISTRY_USERNAME=myusername
 #OCI_REGISTRY_TOKEN=mypassword
 
+# --- Keycloak admin ---
+# Bootstrap credentials for the Keycloak admin console at <host>/kauth.
+# The console is reachable from anywhere the deployment is, so change the
+# password for anything beyond localhost. Seeded on first start only --
+# set it before bringing the stack up, or change it in the console later.
+#KEYCLOAK_ADMIN_USER=admin
+#KEYCLOAK_ADMIN_PASSWORD=change-me
+
 # --- Database passwords ---
 # Each service has its own PostgreSQL instance; passwords can differ.
 # Change these before exposing anything beyond localhost.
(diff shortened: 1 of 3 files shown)


taleodor-claude:
Folded into #235, which now carries both commits (this branch was a descendant of it, so it fast-forwarded cleanly). Closing to keep it to a single PR.

PR #238, opened 2026-07-28
sync(backend): mirror shared backend from rearm-saas — VDR enrichment fix + Team primitive

taleodor-claude:
Mirrors the shared backend tree from rearm-saas via `backend/copy-src.sh`, which by
design leaves `oss/` and `saas/` untouched on both sides.

## What comes across

**1. VDR component enrichment fix** (rearm-saas#372)

VDR export enriches its components from the release's merged SBOM. That enrichment bound
the SBOM with our Jackson 3 mapper, but cyclonedx-core-java is a **Jackson 2** library: it
models `licenses` as a single `LicenseChoice` while the spec serialises it as an ARRAY,
and bridges the two with deserializers registered through Jackson 2 annotations that
Jackson 3 cannot see. It threw on the **first** component carrying licenses, which aborted
the enrichment loop and silently degraded **every** component in the document to the
minimal purl-only fallback — losing licenses, version, hashes and supplier — while the VDR
still exported and still validated, so nothing surfaced to the caller.

Vulnerabilities were never affected: they are built outside the failing block, and the
minimal components keep `bom-ref` set to the PURL so `affects[].ref` still resolves. That
is why it stayed quiet. Live since the Jackson 3 upgrade (2026-05-22).

Now bound through the library's own parser, which carries its own mapper. `parse()` does
not schema-validate, so this is as lenient as the hand-rolled walk it replaces, and an
unparseable BOM degrades to minimal components rather than failing the export.

Note: cyclonedx-core-java **13.0.0 does not fix this** — still Jackson 2, `licenses` still
a singular `LicenseChoice`, and no Jackson 3 migration open upstream.

**2. UserGroup elevated into the Team primitive** (rearm-saas#371) — member roles, external
members, `TeamRole`, and the accompanying schema and tests. This had not been mirrored
yet, so it rides along with this sync rather than being part of the VDR work.

## Verification

- `mvn test-compile` clean against CE's own `oss/` tree — **no OSS reconciliation needed**.
- CE already carries cyclonedx-core-java 12.0.1, so the new parser imports resolve with
  **no dependency change**.
- `VdrComponentBindingTest` passes here 3/3, including a guard rail asserting `Utils.OM`
  still cannot bind cyclonedx model types — so the round-trip is not "simplified" back.
- End-to-end on the SaaS side, the `@vdr_export` harness scenario went red → green against
  a build of this fix.

### DIFF
--- a/backend/src/main/java/io/reliza/model/TeamRole.java
+++ b/backend/src/main/java/io/reliza/model/TeamRole.java
@@ -1,0 +1,30 @@
+/**
+* Copyright Reliza Incorporated. 2019 - 2026. Licensed under the terms of AGPL-3.0-only.
+*/
+
+package io.reliza.model;
+
+/**
+ * The role a member plays on a Team (the {@link UserGroup} primitive).
+ *
+ * <p><strong>A role is an addressing LABEL and never grants access.</strong> It
+ * answers "who on this team do I escalate a KEV finding to?", not "who may write
+ * this component". Membership stays decoupled from RBAC (RFC sec. 4.1): a team
+ * may additionally be granted a permission, but that is a separate, explicit
+ * choice that no role here implies. Nothing in the authorization path reads this
+ * enum -- it feeds notification/escalation targeting only.
+ *
+ * <p>{@code CUSTOM} is the escape hatch for org-defined roles: the operator's
+ * label travels alongside in {@code customRole} rather than widening this enum,
+ * so the known set stays a closed vocabulary (house rule: enums over magic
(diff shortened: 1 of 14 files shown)


taleodor-claude:
Closing in favour of a focused PR carrying only the VDR component enrichment fix.

The full-tree mirror pulled in the UserGroup → Team primitive work (rearm-saas#371) as well, which is still WIP — so it should be synced on its own schedule rather than riding along with a bugfix.

Branch `2026-07-sync-vdr-enrichment-and-teams` is left in place if the Team sync is wanted later.

PR #267, opened 2026-08-06
docs: correct three notification claims that don't match behaviour

logicflakes:
Follow-up to #264. All three of these are corrections to text that PR
introduced, found by walking the page against a running instance instead
of against the source -- which is how they got past review the first time.

## 1. Filters do not work on Community Edition

The page claimed that outside the two Pro channel types and two Pro route
targets, "everything else on this page works the same on either edition."

It does not. `NotificationCelEvaluator` has only a saas-side
implementation. On CE the seam autowires null and the fan-out treats any
non-blank filter as **match-all**:

```java
if (celEvaluator == null) {
    log.debug("No CEL evaluator on this edition; treating non-blank filter "
            + "as match-all for subscription {} on event {}", ...);
    return true;
}
```

That is a deliberate choice -- a subscription that cannot be evaluated
delivers rather than going silent -- but nothing told the operator. A CE
user saves `event.kevListed == true`, sees it accepted, and receives every
event of that type.

Preset toggles are affected too: they compile down to the same expression,
and the gate is whether `celExpression` is blank, not which mode produced
it.

Filtering is now listed as a Pro capability, with a warning spelling out
the CE behaviour and confirming that severity, event types and route
targets *are* still enforced there.

## 2. RELEASE_LIFECYCLE_CHANGED covers 4 of 11 lifecycles

Documented as "a release moves lifecycle stage". Only `DRAFT`,
`ASSEMBLED`, `CANCELLED` and `REJECTED` reach the producer -- an explicit
else-if chain preserving what the older Slack/Teams path notified on.
Subscribing to hear about `READY_TO_SHIP`, `GENERAL_AVAILABILITY` or any
end-of-life stage yields silence, which is the worst way to discover a
gap. The table now names the four and the notable omissions.

## 3. NEW_VULN_AFFECTS_RELEASES can affect nothing

Documented as a vulnerability "newly linked to one of your releases". It
fires for any new vulnerability record in the org; `affectedReleases` is
null at emit and resolved at delivery, and may be empty with no guard.

Verified on a live instance by injecting `CVE-1999-00000`, which no
artifact carries: it enriched to `affectedReleases: 0` and
`affectedComponent: null`, and still fanned out and delivered.

## Checked and left alone

The same pass validated, and found accurate: retention (90 default / 14
floor), dedup (24h default), all eight event types against the enum, the
four route targets, the owner-routing preconditions and fail-closed
behaviour, the CEL limits, both KEV claims (first-sync baseline,
sticky-once-observed), CISA default-on, and the tab locations.

`npm run docs:build` passes with dead-link checking on. Plain ASCII.

### DIFF
--- a/documentation_site/docs/configure/notifications.md
+++ b/documentation_site/docs/configure/notifications.md
@@ -62,6 +62,11 @@ A subscription's `eventTypes` list controls what it can match:
 - Each route on a subscription sets a minimum severity (`CRITICAL` / `HIGH` /
   `MEDIUM` / ...); only events at or above that threshold on that route are
   sent to its targets.
+- A route can also carry a **perspectives** list. Left empty it means "any
+  perspective". Set, it gates delivery: the event only goes out on that route
+  when an affected release's component belongs to one of the named
+  perspectives. This gates **delivery only** -- it does not change what appears
+  in the in-app inbox or the bell.
 
 ::: warning Filters are not applied on Community Edition
 Filter evaluation is a Pro capability. On CE the evaluator is absent, and
--- a/documentation_site/docs/configure/notifications.md
+++ b/documentation_site/docs/configure/notifications.md
@@ -56,9 +56,16 @@ A subscription's `eventTypes` list controls what it can match:
 
 - A subscription can carry an optional filter, built either with the preset
   toggles or as an advanced expression. Filters run under limits (size, nesting
-  depth, iteration count, and a short evaluation timeout) so a runaway
-  expression can't stall delivery for the rest of the org -- a filter that
-  exceeds its budget fails that one delivery rather than blocking others.
+  depth, iteration count, and a 50 ms evaluation budget) so a runaway
+  expression can't stall delivery for the rest of the org. A filter that
+  exceeds its budget -- or fails to evaluate for any other reason -- causes
+  that subscription to be **skipped for that event**; other subscriptions on
+  the same event are unaffected.
+
+  Note that a skip is silent: no delivery row is written, so a filter that
+  never evaluates shows up in [Delivery History](#delivery-history) as
+  *nothing at all*, not as a failure. If a subscription you expect to fire
+  produces no rows whatsoever, an unevaluatable filter is a candidate cause.
 - Each route on a subscription sets a minimum severity (`CRITICAL` / `HIGH` /
   `MEDIUM` / ...); only events at or above that threshold on that route are
   sent to its targets.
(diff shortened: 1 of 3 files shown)


logicflakes:
Heads-up: the `NEW_VULN_AFFECTS_RELEASES` row here needs another pass before this merges — the behaviour it documents is being changed by rearm-saas#396.

This PR currently says:

> The releases it affects are resolved when the event is delivered, and that list can legitimately be empty -- the event still fires, so treat "affects releases" as the intent rather than a guarantee

That was an accurate description of the bug (and finding it is what produced the fix), but rearm-saas#396 makes it false for delivered events: a vuln event whose affected-release set is empty is now withheld rather than delivered, so **a notification you receive always names at least one release**.

The nuance worth keeping is that "empty" was ambiguous — it could mean "nothing is affected" or "the artifact-metrics write that answers the question hasn't landed yet". #396 only suppresses when it can *prove* the metrics are current; when it can't, it retries and ultimately delivers rather than dropping. So the honest post-fix wording is roughly:

> | `NEW_VULN_AFFECTS_RELEASES` | A vulnerability record is created for your organization and at least one of your releases carries it. The affected releases are resolved at delivery time; if nothing in the organization carries the vulnerability, the event is withheld rather than delivered |

Two options, whichever you prefer:
- merge this as-is for the other two corrections (both still accurate and both worth shipping) and I'll follow up with a docs PR once #396 merges; or
- I can push the revised row onto this branch now so it lands in one go.

The other two corrections in this PR — CEL filters not working on CE, and `RELEASE_LIFECYCLE_CHANGED` only emitting for 4 of 11 lifecycle values — are unaffected.

logicflakes:
Update: **two of the three corrections in this PR now document bugs that have since been fixed**, so merging as-is would ship documentation that is wrong in the opposite direction. Only the CE/CEL correction is still accurate.

**1. `NEW_VULN_AFFECTS_RELEASES`** — fixed by rearm-saas#396 (merged `9f585bd2`). An empty affected-release list is no longer delivered.

**2. `RELEASE_LIFECYCLE_CHANGED`** — fixed by your own rearm-saas#395 (merged `ff9e410a`). The gate went from four hardcoded arms to `if (oldLifecycle != curLifecycle)`, so it now fires on every real transition, not just DRAFT/ASSEMBLED/CANCELLED/REJECTED. This PR's row still says "**Only those four**" and calls out `READY_TO_SHIP` / `GENERAL_AVAILABILITY` / end-of-life as emitting nothing — which is exactly what #395 changed.

**3. CEL filters on CE** — unaffected and still correct. `NotificationCelEvaluator`'s only impl is saas-only, the seam autowires null on CE, and `NotificationFanOutService` treats any non-blank filter as match-all. Worth shipping.

Suggested replacement rows:

| Event | Fires when |
|---|---|
| `NEW_VULN_AFFECTS_RELEASES` | A vulnerability record is created for your organization and at least one of your releases carries it. The affected releases are resolved at delivery time; if nothing in the organization carries the vulnerability, the event is withheld rather than delivered |
| `RELEASE_LIFECYCLE_CHANGED` | A release moves from one lifecycle stage to another. Fires on any real transition, including `READY_TO_SHIP`, `GENERAL_AVAILABILITY` and the end-of-life stages |

One nuance you may or may not want in the docs for #396: "empty" was ambiguous between *nothing is affected* and *the artifact-metrics write that answers the question has not landed yet*. The fix only withholds when it can prove the org's metrics are current; when it cannot, it retries and ultimately **delivers** rather than dropping. So the user-facing guarantee is "a notification you receive always names at least one release", without promising the converse.

Happy to push these two rows onto this branch, or open a separate docs PR after this merges — whichever you prefer. Just say which.

PR #268, opened 2026-08-06
rebom: tolerate the enrichment write race in digest-validated BOM fetches

taleodor-claude:
Re-lands the third commit that reached the #263 branch after its merge — same change, verified on the sandbox with the built image (boot clean, raw + processed smoke fetches green), rebom unit suite 165/165.

**The race**: enrichment updates the processed BOM in two steps — overwrite the bytes at tag `<uuid>` in the current-month repository, *then* write the new `processedFileDigest` (+ repo pointer) to the row. A reader whose row snapshot predates step two but whose download lands after step one validates new bytes against the old digest → spurious `Digest validation failed` (observed in production in the current-month repo; nothing corrupt).

**Who hits it — programmatic paths are the primary victims, not user downloads**:
- the **enrichment scheduler** itself: candidate rows are loaded at cycle start, so its expected digest can be *minutes* stale by the time a given BOM is fetched — any concurrent re-push (forced re-enrichment, accept flow, slow prior cycle) trips it;
- the **per-minute SBOM-components reconcile** in the ReARM backend (`parseBomById` → digest-validated fetch): fires for every fresh upload while that upload's first enrichment is still in flight — every new upload races its own enrichment;
- DTrack synthetic-bucket building / enriched-BOM probes on their schedules.
User-initiated downloads/exports read row-and-fetch back-to-back (milliseconds window) — possible colliders, but the least likely ones.

**Fix**: `fetchProcessedBomWithRetry` wraps every digest-validated processed-BOM read (the `findBomObjectById` family, version/format paths, the scheduler fetch, the raw-resolver substitute). On a `DigestValidationError` it re-reads the row **once**: digest/repo changed under the reader ⇒ it was the race, refetch against fresh values; row unchanged ⇒ genuine corruption, the *original* error rethrows untouched. Non-digest errors pass through; the happy path adds nothing.

Tests: 6 unit cases (race retry with fresh digest+repo, unchanged-row corruption rethrow, deleted-row rethrow, one-shot limit, non-digest passthrough, clean-path no-re-read).

### DIFF
--- a/rebom-backend/src/services/bom/bomCrudService.ts
+++ b/rebom-backend/src/services/bom/bomCrudService.ts
@@ -3,7 +3,7 @@ import { BomDto, BomMetaDto, BomRecord, BomSearch, SearchObject } from '../../ty
 import { BomNotFoundError, BomDataIntegrityError } from '../../types/errors';
 import { BomMapper } from './bomMapper';
 import { logger } from '../../logger';
-import { fetchFromOci, extractRepositoryNameFromBom, extractRepositoryNameFromSpdxOciResponse, resolveAndFetchRawBom } from '../oci';
+import { fetchFromOci, extractRepositoryNameFromBom, extractRepositoryNameFromSpdxOciResponse, resolveAndFetchRawBom, fetchProcessedBomWithRetry } from '../oci';
 import { parseBom, ParsedBom, ParsedBomComponent, parseHbom, ParsedHbom } from './bomComponentExtractor';
 
 export async function findAllBoms(): Promise<BomDto[]> {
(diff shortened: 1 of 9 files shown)


taleodor-claude:
Review follow-up (92286663): three digest-validated processed fetches had been missed by the original sweep and were still on the raw `fetchFromOci` path — `triggerEnrichment` non-forced (highest exposure: a manual trigger most likely collides with another enrichment finishing), `findBomByMeta`, and `BomMapper.toDtoWithContent` (currently uncalled; migrated so a future caller cannot reintroduce the race). All `processedFileDigest` readers now go through `fetchProcessedBomWithRetry`. The SPDX-source fetch at bomProcessingService.ts:1068 intentionally stays raw — it validates a write-once artifact with its own digest field, outside the race class. tsc clean, unit suite 165/165.

PR #271, opened 2026-08-11
fix(ui): delivery-status filters didn't match what the backend writes, in both directions

logicflakes:
Swept every write path for `notification_deliveries.status` — `setStatus()` call sites, the JPA field default, native `UPDATE`s, migrations. **The backend produces exactly `PENDING`, `SENT`, `FAILED`, `BATCHED`.**

The filter offered nine options and got it wrong in both directions:

| Status | Written? | Was | Now |
|---|---|---|---|
| PENDING / SENT / FAILED | yes | selectable | selectable |
| **BATCHED** | **yes** | **missing entirely** | **added, selectable** |
| **ACKED** | **no** | selectable | disabled |
| RATE_LIMITED | no | selectable | disabled |
| EVAL_TIMEOUT | no | selectable | disabled |
| TEST | no | selectable | disabled |
| PREVIEW | no | selectable | disabled |

## Why the dead ones matter

Same silent trap #197 just closed for the PREVIEW *subscription* status: filtering by one returns zero rows forever, which reads as a fact about your data rather than about the schema.

**`ACKED` is the one that bites.** An operator marks 40 inbox items read, filters Delivery History by `ACKED` to see what was acknowledged, and gets nothing — because `setAckedAt()` has **zero callers** and read state is a `LEFT JOIN` on `notification_reads`, not a status. The others: `RATE_LIMITED` (limit stored, not enforced), `EVAL_TIMEOUT` (failed filters are skipped and logged, no row written), `TEST` (superseded by the ORIGIN dimension), `PREVIEW` (mode never implemented).

## Why BATCHED matters

Produced since the email-digest work, present in the GraphQL filter enum, and absent from the UI — so **email-digest deliveries could not be filtered for at all**. We watched `BATCHED` rows land during the notifications walkthrough; there was no way to query them.

## Treatment

Dead statuses are **visible-but-disabled**, labelled `(not yet available)`, matching #197 and `VEX_STATE_CHANGED` so the reason is on screen rather than the option quietly vanishing. **Backend enums untouched** — they're the design intent, and removing them would break the GraphQL API for any client that sends one.

## The guard

The spec pins **both directions** against the backend source: it scrapes `setStatus()` call sites and asserts `disabled ⟺ no writer`, plus that every produced status has an option.

It scans the **CE mirror always and Pro via `it.runIf`**, because the mirror lacks the `service/saas` classes where a new writer is most likely to land — scanning only CE would leave the guard unable to fire for the case it advertises (a Pro-only `EVAL_TIMEOUT` writer).

**Mutation-verified three ways:** enabling a writerless status, disabling a written one, and removing a written one entirely each fail it.

## One correction worth recording

An earlier cut of this exempted `ACKED` via a hardcoded whitelist, on the assumption it was written by the read-tracking path. It isn't — and that whitelist was exactly what kept the test green over a fifth dead status. Caught in review. The produced set is now **derived from the source**, never asserted by hand.

Browser-verified against the sandbox: all five dead statuses render disabled, all four live ones selectable.

138 tests pass, lint clean, build clean.

Part of `t20260807-151031-14455`.

### DIFF
--- a/ui/src/utils/notificationsCommon.spec.ts
+++ b/ui/src/utils/notificationsCommon.spec.ts
@@ -8,13 +8,14 @@ vi.mock('@/utils/commonFunctions', () => ({
     default: { dateDisplay: (s: string) => `ABS:${s}` },
 }))
 
-import { readFileSync, existsSync } from 'fs'
+import { readFileSync, existsSync, readdirSync } from 'fs'
 import { fileURLToPath } from 'url'
+import { join } from 'path'
 
 import { print } from 'graphql'
 import {
     relativeTime,
-    subscriptionStatusOptions, eventTypeOptions,
+    subscriptionStatusOptions, eventTypeOptions, deliveryStatusOptions,
     buildNotificationFilterInput,
     routeHasTarget,
     classifySubscriptionTest,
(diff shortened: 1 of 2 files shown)


logicflakes:
## Two-layer review run on the settled diff — three corrections

I'd previously reviewed an earlier revision of this PR and then changed a lot in response, so the gate has now been re-run properly as two independent passes against what's actually here. Both found real things.

### The one that mattered: the guard could become unsatisfiable

Asserting `selectable == produced` against **both** the CE mirror and Pro breaks the moment a writer lands in `service/saas`, which CE doesn't mirror: Pro demands the option enabled, CE demands it disabled, and **no value of the option list passes both**. The only escape is deleting a test — the failure mode where a guard gets disarmed rather than obeyed.

Fixed by asserting *different* propositions per tree, matching what `notificationInboxSchemaDrift.spec.ts` already does for the same reason: **Pro (source of truth) gets the exact both-ways assertion; CE gets containment only.**

Proven by planting a Pro-only writer in `service/saas`:

```
before  Pro: RATE_LIMITED writer=true  -> expected true to be false
        CE:  RATE_LIMITED writer=false -> expected false to be true    (no green state)
after   Pro test alone fails; enabling the option makes BOTH green
```

### The guard was blind to a writer that already exists

It matched only `setStatus(...)`, so the native-SQL writer in `NotificationDeliveryRepository` was invisible — its coverage was complete only by the coincidence that that writer happens to write `PENDING`, which the same revision hardcoded. Now both idioms are matched, in the **assigning** position specifically:

```
setter-form hits in that file : (none)
native-SQL WRITE hits         : [ PENDING ]
all _VALUE mentions in file   : PENDING, SENT, ACKED, BATCHED
```

The three `WHERE`-clause `_VALUE` references are correctly ignored. `PENDING` is now derived rather than asserted.

### Comments are stripped before scanning

Javadoc here quotes call forms — `NotificationCelEvaluatorImpl`'s class doc names `NotificationDeliveryStatus.EVAL_TIMEOUT` *precisely while explaining that nothing writes it*. Reading prose as code would have demanded we enable that filter.

### From the conventions pass

`_PATH` suffix on the path consts; CE/Pro split into separate describes with honest titles so a CE-only run reads truthfully in the reporter; module block moved so it stopped hijacking #197's comment; option array typed so the `as { disabled?: boolean }` cast disappears. Plus a real catch: **`BATCHED` was added to the option list but not to `deliveryStatusTagType`**, so a produced status rendered with a default tag while the *disabled* ones got colours.

### Verification

Mutation-verified four ways — enabling a writerless status, disabling a written one, dropping a written one, and dropping the natively-written one each turn it red. Browser-re-probed against the sandbox after the rebuild: all five dead statuses disabled, all four live ones selectable. 138 tests, lint clean, build clean.

### One thing this PR does not fix

**These specs never run in CI.** `ui/package.json` defines `test: vitest run` and the workflow has no vitest step, so every drift guard in `ui/src/utils` — including #197's PREVIEW tripwire and this one — is developer-box-only. Filed as `t20260811-061400-10264` rather than scope-creeping here.

PR #273, opened 2026-08-11
feat(ui): drop the owner suggestion and the Component Leads editor

logicflakes:
Two removals from the component settings form, both from rhythm's manual UI surface review.

## 1. The owner candidate suggestion

The Owner block printed the backend's candidate-team suggestion as the reason line under an `UNSET` status:

> **UNSET** — No owner set; suggest team 'Platform Team'

Sitting directly above the owner picker, a suggestion reads like a decision that has already been made. The `UNSET` tag on its own says what is true.

**Only the `UNSET` reason is suppressed.** `DEGRADED` ("owner team is archived"), `ORPHANED` and `NON_DURABLE` still explain themselves — those are statements about a real owner, not a guess at one.

| before | after |
|---|---|
| `UNSET` + "No owner set; suggest team 'Platform Team'" | `UNSET` |

## 2. Component Leads

Leads were the pre-ownership way of naming who is accountable for a component, and Owner supersedes them — the field's own backend javadoc already claims leads are "retired into" owner (they were not; this finishes the UI half). Two overlapping answers to one question are worse than one.

The backend keeps the field and any stored value survives. This form no longer sends `leads` at all, and an omitted value reads as "leave unchanged". Same shape as #197.

## Validation

Hot-swapped onto the sandbox and probed in a real browser (`probe-owner-suggestion-and-leads.js`, `probe-leads-not-clobbered.js`).

**The suggestion check would have passed vacuously**, because no Demo org group held a component permission and so no component produced a suggestion at all. Setup granted one first, and the probe was run against the pre-change build to confirm it *fails* there — baseline shows "suggest team 'Platform Team'" and the Leads editor; after, both gone, `UNSET` and Stakeholder Contacts intact. 6/6 checks.

**The no-clobber check nearly went vacuous too.** The first target component (Automation Tutorial Vue) rejects every save on a dangling `vcs` reference in the demo data, so "stored leads survived" was true because nothing was written. The probe now asserts the backend *accepted* the save. On a component that saves cleanly: lead seeded, two accepted saves (revision 0 → 2), value unchanged, and the captured mutation payload carries no `leads` key.

`npm run test` 138/138, `npm run build` clean, `validate-graphql.mjs` 306 documents / 0 failures. Lint is 273 errors on this branch and 273 on `main` — pre-existing, unchanged.

Sandbox data restored (probe permission revoked, seeded leads removed, `kubectl-set` field manager stripped).

### DIFF
--- a/ui/src/components/ComponentView.vue
+++ b/ui/src/components/ComponentView.vue
@@ -405,12 +405,10 @@
                                                         {{ componentOwnership.ownership.status }}
                                                     </n-tag>
                                                     <span v-if="ownerLabel" style="margin-left: 8px;">{{ ownerLabel }}</span>
-                                                    <!-- UNSET's reason is the backend's candidate-team SUGGESTION
-                                                         ("suggest team 'X'", "N candidate teams", "create one").
-                                                         Suppressed: nobody asked for a suggestion here, and offering
-                                                         one beside a picker invites reading it as a decision that has
-                                                         already been made. The UNSET tag alone says what is true. -->
-                                                    <span v-if="componentOwnership.ownership.reason && componentOwnership.ownership.status !== 'UNSET'"
+                                                    <!-- A suggestion's reason line is suppressed: offering a candidate
+                                                         team directly above the owner picker invites reading it as a
+                                                         decision already made. Every other status explains itself. -->
+                                                    <span v-if="componentOwnership.ownership.reason && !ownershipIsSuggestion"
                                                         class="text-muted" style="display: block; margin-top: 4px;">
                                                         {{ componentOwnership.ownership.reason }}
                                                     </span>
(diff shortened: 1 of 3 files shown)


logicflakes:
## Both review layers run against the settled diff

**Layer 1** raised two things and I applied one:

- **The suppression is broader than "the suggestion".** `suggestFromCandidates` emits three `UNSET` reasons; the third — *"N individual write-team member(s) but no team — create one"* — proposes no team, it explains why none can be suggested. It is now hidden with the others. Left as-is (it is noise without ownership governance in use), but flagging it as a decision rather than collateral. If you want it back, key the suppression on `ownership.ownerRef != null` instead of on the status.
- **`UNSET` now renders as a bare enum token.** Every old reason began "No owner set"; now the chip stands alone next to a helper line about durability. Not fixed — say the word and I will add a fixed `No owner set.` line, which states the fact without proposing anything.

**Layer 2** found three real ones, all fixed in `54f23dea`:

- **A third copy of the `'UNSET'` string literal.** The template `v-if` and `ownerLabel` *must* agree — if they drift, the UI prints a suggested team beside a tag saying nothing is set. Now one `ownershipIsSuggestion` computed, read by both.
- **The `store.ts` comment had drifted into being false.** Deleting `leadDetails` from its enumeration left *"contacts is the only writeable one left"* floating above twenty writeable keys. Scoped to the people fields.
- **My tombstone comment overstated the change** — see below.

## One thing worth your attention before merging

My comment said Owner "supersedes" Leads. **It does not, yet.** There is no leads-to-owner backfill anywhere in the backend — `ComponentData.owner`'s javadoc claims `leads` "is retired into this field on backfill", but no migration and no service implements it, and `candidateOwnerTeams` derives suggestions from `READ_WRITE` grants and never consults `leads`.

So for any component whose only accountability record is `leads`, this PR removes the last place that information was visible. The data is preserved and readable over the API — nothing is lost — but there is no UI left to see it and nothing moves it into Owner.

That may be fine given ownership is new and leads may be unused too. But you told me ownership specifically is not in use, which makes `leads` the *older* and more likely-populated field of the two. If any org has leads set, a `leads` → `owner` backfill probably wants to land before or with this. Comment now says what is true either way.

Probe extended to cover what the refactor could have broken: an owned component must still name its owner and still render a non-suggestion reason. 9/9 on the sandbox, 138/138 unit tests, build and `validate-graphql` clean.

PR #274, opened 2026-08-11
feat(ui): withdraw the External Members editor -- nothing ever notified them

logicflakes:
The team editor told operators:

> People reachable only by email or handle, such as a vendor contact. **They can be notified**, but do not count toward the team's durability.

They cannot be notified. The field is collected, stored, rendered back — and no code path has ever sent anything to one of those contacts.

## Why hiding, and not delivering

Rhythm's call was "deliver to them, or hide until we have the mechanism". There is no mechanism, and building one is a feature with its own design decisions. Three separate things are missing:

1. **No transport discriminator.** `contact` is one freeform string. Nothing distinguishes an email address from a Slack handle, so fan-out cannot pick a dispatcher.
2. **Email cannot address a person.** `EmailChannelDispatcher` reads a *static* recipient list off the channel. There is no per-delivery override, so adding an external member to one event's send would mean editing the channel itself.
3. **Slack and Teams cannot DM at all.** Both post to an incoming-webhook URL, which delivers to one preconfigured channel. DMing a handle needs a bot token and `chat.postMessage` — a different integration with its own scopes, install flow and secret handling.

So a handle-shaped contact is undeliverable today at any price. Filed as its own task; the Team RFC deliberately does not carry `externalMembers` forward, so the delivery feature gets to choose its own shape — most likely a typed contact rather than a free string.

## Stored data is untouched

The form no longer sends `externalMembers` at all, and the backend reads an omitted field as "keep" — sending `[]` would REPLACE, i.e. delete.

**Verified rather than assumed:** a group carrying two real external members went through two *accepted* saves (the probe asserts the backend accepted them, so this can't pass because nothing was written) and kept all three of its team fields — external members 2, member roles 1, channels 2, name restored.

## Helpers removed too

`buildExternalMembersPayload` and the externals arm of `validateTeamMembers` go with the editor rather than lingering as unreachable code. The returning feature needs a typed contact, so these builders would not have fitted it.

**Unit suite goes 142 → 138** — 5 tests removed with the code they covered, 1 added for the now-empty validation case. Build and `validate-graphql` clean.

## Validation

`probe-external-members-hidden.js`, 7/7 against the sandbox. It sanity-checks that the modal actually opened before asserting the section is absent, and restores the group name it edited.

### DIFF
--- a/ui/src/components/OrgSettings.vue
+++ b/ui/src/components/OrgSettings.vue
@@ -620,35 +620,12 @@
                                 </div>
                             </n-space>
 
-                            <n-space v-if="teamMembersKnown && !restoreMode" vertical style="margin-bottom: 20px; width: 100%;">
-                                <n-h5>
-                                    <n-text depth="1">
-                                        External Members (no ReARM account):
-                                    </n-text>
-                                </n-h5>
-                                <n-text depth="3" style="font-size: 12px;">
-                                    People reachable only by email or handle, such as a vendor contact. They can be
-                                    notified, but do not count toward the team's durability.
-                                </n-text>
-                                <n-dynamic-input
-                                    v-model:value="teamExternalMembers"
-                                    :on-create="onAddExternalMember"
-                                >
-                                    <template #default="{ value }">
-                                        <div style="display: flex; gap: 8px; width: 100%;">
-                                            <n-input v-model:value="value.name"
-                                                placeholder="Name" style="width: 200px;" />
-                                            <n-input v-model:value="value.contact"
-                                                placeholder="Email or handle" style="width: 240px;" />
-                                            <n-select v-model:value="value.role" :options="constants.TeamRoles"
-                                                placeholder="No role" clearable style="width: 200px;" />
-                                            <n-input v-if="value.role === constants.TeamRoleCustom"
-                                                v-model:value="value.customRole"
-                                                placeholder="Custom role label" style="width: 200px;" />
-                                        </div>
-                                    </template>
-                                </n-dynamic-input>
-                            </n-space>
+                            <!-- The External Members editor lived here. It promised "they can be
+                                 notified" and nothing ever delivered to an external contact: there is
+                                 no transport discriminator on the contact string, the email dispatcher
+                                 sends to a static per-channel recipient list, and Slack/Teams post to
+                                 an incoming webhook that cannot DM anyone. Stored values are kept and
+                                 are not sent by this form. -->
 
                             <n-space v-if="teamMembersKnown && !restoreMode" vertical style="margin-bottom: 20px; width: 100%;">
                                 <n-h5>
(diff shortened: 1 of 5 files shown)


logicflakes:
## Layer 2 outcomes (`e656d9c9`)

**The one it would have blocked on, and it was right.** The entire safety property of this PR is that the mutation never carries an `externalMembers` key — and it was protected by a *comment*. Since nothing in the UI shows external members any more, a later refactor re-adding `externalMembers: []` would destroy every team's stored contacts with no failing test and nobody watching.

The payload assembly is now `utils/userGroupUpdateInput.ts` with a spec. Two neighbouring rules came along, both the same omit-or-you-delete shape and both previously untested:

- team fields are omitted until this group's data has actually loaded — sending `[]` from an unloaded editor wipes the roles;
- channels are omitted when the channel list failed to load — a partial picker must not overwrite the real list.

**Suite 138 → 143** (so the branch now sits above `main`'s 142 rather than below it).

Also fixed: three stale comments still claiming external members can be stripped by a save that can no longer touch them; a comment in `teamMembers.ts` contradicting the mirror it describes; and the template tombstone, which asserted backend dispatcher internals from a Vue template — accurate today, guaranteed to rot. Now one line pointing at the util.

It separately verified the premise I had asserted: omission really does preserve, via `UserGroupDataFetcher` → `convertValue` → null → the explicit `null` check in `UserGroupData:404`, with the schema field nullable and an existing backend test already pinning it.

Re-probed after the extraction: 7/7, and both Demo teams kept every external member, role and channel across two accepted saves.

## One thing I want a decision on

Layer 2 raised a data-protection angle I had not considered, and I think it is right:

> Stored external contacts are now permanently uneditable and undeletable from the UI... `contact` is a freeform email address (PII) that operators now have no way to view, correct, or erase.

There is no other surface — `rearm-cli` has no `externalMembers` support either. So "kept forever, invisible" is arguably a worse posture than either editable or cleared, even though preserving is the right default for a *hide*.

Both Demo teams have real addresses stored right now, which is a fair sample of what a customer would have. Options:

1. **Ship as-is**, and let the delivery feature (which restores an editor) be the answer. Simplest; leaves invisible PII in the meantime.
2. **Keep a read-only display** of stored contacts with no editor. The false "they can be notified" claim goes, visibility stays. Small addition.
3. **Clear stored external members** as part of this change. Cleanest posture, destroys data an operator entered deliberately.

I lean 2 — the actual defect was the false promise, not the visibility, and it keeps the erase path open via the existing editor semantics. Happy to ship 1 if you would rather not grow the change. Not doing 3 without an explicit call.

logicflakes:
**Decision (rhythm, 2026-08-11): ship as-is.** Stored external contacts stay in the database, unreachable from any UI, until the delivery feature restores an editor. No read-only display, no purge.

Recorded on the delivery task (`t20260811-115822-32525`) so whoever picks it up knows there is existing data to surface, not a clean slate — and that an operator has had no way to correct or erase it in the meantime.

No further changes to this PR; it is ready to merge.

PR #276, opened 2026-08-11
feat(ui): Teams tab -- create and edit the org's addressable unit (Phase 1)

logicflakes:
UI half of Phase 1. Needs the backend in rearm-saas#409 (Pro schema) to be deployed; degrades cleanly without it.

A new org-admin **Teams** tab beside User Groups. Create a team, then set its members, contained user groups, notification channels and leads. Nothing else on the page changes behaviour.

## Two things the form has to get right

**The leads picker mirrors the backend rule.** `TeamService` requires a lead to be on the effective roster — which includes people who are on the team only through a *contained user group*. So the picker resolves the same way, and a watcher strips leadership when someone leaves the roster. Without both, the form cheerfully assembles a save the backend rejects with no visible cause. Same pre-submit-mirror approach the existing team-role editor uses.

**Archive/restore, not delete.** A team is a reference target; deleting the row would strand pointers. Archiving keeps the roster and reserves the name, and is reversible.

## Drift isolation

Teams is Pro-ahead of the CE mirror, so the pane runs its own query and hides itself on a schema-drift error rather than taking the page down with it.

## A hole I dug and then closed

I first factored the three Team selections into a shared `${TEAM_FIELDS}` constant. `scripts/validate-graphql.mjs` **skips any gql template containing an interpolation** — it can't statically parse one — so that quietly opted the entire Team surface out of the only check that catches schema drift. The run reported a clean `306 documents, 0 warnings` and meant nothing.

Inlined, the validator now reports **0 Pro failures and 3 CE-mirror warnings**, which is correct: Team doesn't exist on the CE mirror yet, and the WARN category is exactly "valid on Pro, must degrade gracefully at runtime" — which the pane does. A comment records why the repetition is deliberate, so nobody tidies it back out of coverage.

## Checks

`npm run build` clean, 143/143 unit tests, `validate-graphql` as above.

Browser validation against the sandbox needs #409 deployed there first; I'll probe it once that lands, rather than claiming a check I haven't run.

### DIFF
--- a/ui/src/components/OrgSettings.vue
+++ b/ui/src/components/OrgSettings.vue
@@ -679,6 +679,12 @@
                 </div>
             </n-tab-pane>
 
+            <n-tab-pane v-if="isOrgAdmin" name="teams" tab="Teams">
+                <TeamsOfOrg
+                    :orguuid="orgResolved"
+                    :isWritable="isWritable" />
+            </n-tab-pane>
+
             <n-tab-pane name="programmaticAccess" tab="Programmatic Access" v-if="isOrgAdmin">
                 <div class="programmaticAccessBlock mt-4">
                     <h5>Programmatic Access</h5>
@@ -1218,6 +1224,7 @@ import ScopedPermissions from './ScopedPermissions.vue'
 import OrgIntegrations from './OrgIntegrations.vue'
 import OrgGlobalApprovalPolicyRules from './OrgGlobalApprovalPolicyRules.vue'
 import OrgTeamAssignmentRules from './OrgTeamAssignmentRules.vue'
+import TeamsOfOrg from './TeamsOfOrg.vue'
 import AiAgentPoliciesOfOrg from './AiAgentPoliciesOfOrg.vue'
 import CommittersOfOrg from './CommittersOfOrg.vue'
 import { FetchPolicy } from '@apollo/client'
--- a/ui/src/components/TeamsOfOrg.vue
+++ b/ui/src/components/TeamsOfOrg.vue
@@ -1,0 +1,467 @@
+<template>
+    <div class="teams-pane">
+        <div class="tab-toolbar">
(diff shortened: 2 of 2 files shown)


logicflakes:
## Validated in a real browser against the sandbox

Both images (this branch + rearm-saas#409) deployed to the claude2 sandbox. `probe-teams-phase1.js`, **11/11**:

```
PASS  the Teams tab rendered against a backend that has Teams
PASS  the teams table is present
PASS  the team appears in the list after create
PASS  a duplicate name is rejected (got: "A team named 'PROBE Team Phase1' already exists.")
PASS  leads picker is EMPTY on a team with no roster (got 0)
PASS  attaching "Docs Team" makes its members eligible leads (got 1: Kev Probe (…))
PASS  a group-derived member can be set as lead
PASS  a notification channel can be attached (picked: Walkthrough Email (Email))
PASS  the save was accepted
PASS  archive marks the team INACTIVE
PASS  restore brings it back to ACTIVE
```

**The two roster checks are the ones that matter, and they're a matched pair.** The picker is empty on a team with no roster, then attaching *Docs Team* makes its member eligible — so the containment rule is demonstrated rather than assumed, and neither check can pass vacuously: if the picker offered all org users it would fail the first, and if it ignored contained groups it would fail the second.

The persisted row confirms it end to end — `members` empty, `leads` holding a user who is on the team **only** through the attached group.

The unsupported-backend path is also confirmed negatively: `teams-unsupported` is absent here, so the pane really did talk to a backend that has Teams rather than silently degrading and passing everything else.

Separately, `probe-teams-backend-rules.js` (on the backend PR) drives the API directly with a lifted bearer token to prove these rules hold with the UI bypassed — the pickers are a convenience, not the guard.

Sandbox restored: probe teams and audit rows deleted, Demo groups untouched, field managers stripped. Both images left running so the tab is browsable.

PR #277, opened 2026-08-11
fix(ui): a double-click on Trigger Auto Integrate sent two mutations

logicflakes:
## The problem

`autoIntegrateFeatureSet` is **not idempotent**. Server-side it gathers each dependency's latest release, asks whether a matching product release already exists, and creates one if not — a check-then-act that takes no lock, so two in-flight calls can both find nothing and both create a product release.

The control is a plain `n-icon` with an `@click` handler: no built-in disabled or loading state, and the handler had no re-entry guard. A double-click fired two concurrent mutations.

## The fix

Guards on an in-flight ref and reflects it in the control — reduced opacity, `not-allowed` cursor, and the tooltip switching to *"Auto Integrate in progress..."*.

Reset in a `finally`, **not** after the notify, so a failed mutation cannot leave the button permanently dead.

## Why this instead of a server-side lock

The server-side fix was implemented and measured (relizaio/rearm-saas#408, now **closed**): four concurrent calls did produce four product releases. But the evidence did not justify shipping it:

- **Zero** feature sets in sandbox data have two product releases sharing a parent set — the path being unguarded had never actually caused harm. The 4/4 number came from a synthetic four-thread harness, not a realistic scenario.
- The lock would have been held across `createProductRelease -> createRelease`, which performs **synchronous external I/O** (GitHub build trigger, email send). A stalled user request would then block the *automatic* integrator for that feature set until a 30s lock timeout — a failure mode that does not exist today.

Trading a real new failure mode for a theoretical one was the wrong direction. A double-click is the realistic trigger, and this covers it with no locking, no new coupling, and no risk to the version pipeline.

## Checks

- UI build clean.
- **No new lint violations** — the repo's error count is unchanged from main at 273 (pre-existing, repo-wide). My first attempt added 13 by wrapping the body in `try` without re-indenting; fixed.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/ui/src/components/BranchView.vue
+++ b/ui/src/components/BranchView.vue
@@ -181,8 +181,13 @@
                         @click="showAddComponentModal = true" title="Add Dependency" size="20" style="margin-left: 4px; vertical-align: middle;">
                         <CirclePlus />
                     </n-icon>
-                    <n-icon v-if="isWritable && branchData.autoIntegrate === 'ENABLED' && modifiedBranch.autoIntegrate === 'ENABLED'" class="clickable"
-                        @click="triggerAutoIntegrate" title="Trigger Auto Integrate" size="20" style="margin-left: 4px; vertical-align: middle;">
+                    <n-icon v-if="isWritable && branchData.autoIntegrate === 'ENABLED' && modifiedBranch.autoIntegrate === 'ENABLED'" :class="autoIntegrateInFlight ? '' : 'clickable'"
+                        @click="triggerAutoIntegrate"
+                        :title="autoIntegrateInFlight ? 'Auto Integrate in progress...' : 'Trigger Auto Integrate'"
+                        size="20"
+                        :style="{ 'margin-left': '4px', 'vertical-align': 'middle',
+                                  opacity: autoIntegrateInFlight ? 0.5 : 1,
+                                  cursor: autoIntegrateInFlight ? 'not-allowed' : undefined }">
                         <TrendingUp />
                     </n-icon>
                 </p>
@@ -449,6 +454,7 @@ const selectNewVcsRepo = ref(false)
 
 
 const compareMode = ref(false)
+const autoIntegrateInFlight = ref(false)
 const comparisonCheckboxes: Ref<any> = ref({})
 const showReleaseComparisonModal = ref(false)
 
@@ -566,24 +572,36 @@ const createFsFromRelease = async function(){
 }
 
 async function triggerAutoIntegrate () {
-    const resp = await graphqlClient.mutate({
-        mutation: gql`
-            mutation autoIntegrateFeatureSet($branchUuid: ID!) {
-                autoIntegrateFeatureSet(branchUuid: $branchUuid) {
-                    uuid
-                    version
-                }
-            }`,
-        variables: { branchUuid: branchUuid.value },
-        fetchPolicy: 'no-cache'
-    })
-    const release = (resp.data as any)?.autoIntegrateFeatureSet
-    await onCreated()
-    if (release && release.version) {
-        notify('success', 'Auto Integrate Completed', `Release ${release.version} was created via auto-integration.`)
-        showBranchSettingsModal.value = false
-    } else {
-        notify('info', 'Auto Integrate Completed', 'Auto-integrate was attempted, but no new release was created.')
+    // The mutation is not idempotent: autoIntegrateFeatureSetOnDemand gathers each dependency's
+    // latest release, checks whether a matching product release already exists, and creates one if
+    // not. That check-then-act takes no lock server-side, so two in-flight calls can both find
+    // nothing and both create a product release. The button is a plain icon with no built-in
+    // disabled state, so without this a double-click sends two mutations.
+    if (autoIntegrateInFlight.value) return
+    autoIntegrateInFlight.value = true
+    try {
+        const resp = await graphqlClient.mutate({
+            mutation: gql`
+                mutation autoIntegrateFeatureSet($branchUuid: ID!) {
+                    autoIntegrateFeatureSet(branchUuid: $branchUuid) {
+                        uuid
+                        version
+                    }
+                }`,
+            variables: { branchUuid: branchUuid.value },
+            fetchPolicy: 'no-cache'
+        })
+        const release = (resp.data as any)?.autoIntegrateFeatureSet
+        await onCreated()
+        if (release && release.version) {
+            notify('success', 'Auto Integrate Completed', `Release ${release.version} was created via auto-integration.`)
+            showBranchSettingsModal.value = false
+        } else {
+            notify('info', 'Auto Integrate Completed', 'Auto-integrate was attempted, but no new release was created.')
+        }
+    } finally {
+        // finally, not after the notify: a failed mutation must not leave the button dead.
+        autoIntegrateInFlight.value = false
     }
 }
 


logicflakes:
## Reviewed (it never had been), findings fixed in `41a7067f`

Also merged `origin/main` -- the branch was 72 commits behind, clean merge, no
conflicts.

**1. `try/finally` with no `catch`.** A rejected mutation reported nothing: the
global handler in `router.ts` only reacts to stale-chunk errors, so the operator
saw silence and an unrefreshed view. Now uses this file's own house pattern
(`releaseLock`, `saveBranch` and `fetchComponentBranches` all do it):
`notify('error', ...)` with `extractGraphQLErrorMessage`.

The catch also refreshes, which is the same argument the guard itself rests on:
because the mutation is not idempotent, a failure does **not** mean nothing
happened -- a timeout after the release was committed is indistinguishable from
one before it -- and an operator who reads an error and clicks again is exactly
how the second product release gets made. The refresh is best-effort
(`.catch(() => {})`) because `onCreated` fetches too; whatever broke the mutation
has usually broken it as well, and a throw inside the catch would swallow the
error report and escape as an unhandled rejection, reintroducing the silence
this commit removes in the case it matters most.

**2. The race claim was overstated.** The comment presented the unlocked
server-side check-then-act as the reason for the guard, which reads as though
the guard closes it. It does not -- the ref is per component instance, so two
tabs, two operators, or any API caller still produce two concurrent mutations
and two product releases. The comment now says what the guard does and does not
do, and names what a real fix needs: a lock or an idempotency key server-side.
**Worth a follow-up issue on `rearm-saas`** -- happy to open one.

The guard itself came through review clean: flag set before the `await`, reset
in `finally`, early return ahead of the mutation, no other caller of
`triggerAutoIntegrate`.

## Validation

- `npm run build` clean.
- `npm test`: **4 specs fail, none of them this file** -- all CE-schema-drift
  checks (`routeInputSchemaDrift`, `instanceFragmentsSchemaDrift`,
  `releaseFragmentsSchemaDrift`). I ran them on a clean `origin/main` checkout
  and got the identical 4 failures, so **main is red independently of this
  branch**. Notably the retarget for one of them ("CE now accepts
  notifyComponentOwner") is sitting unmerged in #325.

logicflakes:
## Re-reviewed the review fixes -- and that is where the real bug was (`89a0ed53`)

The first review's findings got fixed, but the fixes themselves had not been
reviewed. Re-running the pass on just that delta found two defects **I
introduced yesterday**, one of them worse than the finding it came from.

**1. The catch wrapped too much, so a success could be reported as a failure.**
`await onCreated()` sat inside the new `try`, which means a refresh that failed
*after* the mutation had already created the release was reported as **"Auto
Integrate Failed"**. That is the most harmful message available here: the
release exists, the operator is told it does not, and clicking again is the
reasonable response -- manufacturing precisely the duplicate product release
this PR exists to prevent. The mutation now sits in its own `try`; past that
point the outcome is reported either way and the refresh is best-effort.

**2. The error-path refresh silently reset the modal's edit form.** I used
`onCreated()`, which reassigns `modifiedBranch` from `branchData`. The success
path can afford that because it closes the modal immediately after -- the error
path leaves it open, so unsaved branch-settings edits vanished with no prompt,
bypassing the confirm that `handleBranchSettingsClose` exists to give. It now
dispatches `fetchReleases`, which feeds the `releasesOfBranch` getter the list
renders from, and touches nothing else.

**3. Narrowed an overclaim.** The refresh does not put the release in front of
the operator -- the list renders behind the modal, which stays open. What it
does is make the list correct by the time the modal closes, rather than stale
until something else fetches. The comment says that now.

### Not fixed -- pre-existing, flagged rather than silently changed

When the mutation **succeeds but creates no release**, the modal stays open and
`onCreated()` has already reset the form: the same silent loss of unsaved edits
as #2, on a path this PR neither introduced nor otherwise touches. Fixing it
means deciding what that path should do (leave the form alone? close the modal?
warn?), which is a different question from the double-click. Happy to take it as
a follow-up.

### Validation

- `npm run build` clean.
- `npm test` still fails the same 4 CE-schema-drift specs, verified identical on
  a clean `origin/main` checkout -- unrelated to this branch.
- Confirmed by review and by reading: the in-flight flag cannot leak (reset in
  `finally` on every path, including a throwing `notify`), `fetchReleases` is a
  registered async Vuex action so `.catch()` on the dispatch is safe, and the
  error path leaves `modifiedBranch` / `originalBranch` / `hasBranchChanges`
  untouched so the close guard still works.

logicflakes:
## Verified on the sandbox -- reproduced the bug, then proved the fix

Not "it builds". A/B on a live sandbox with a purpose-built fixture: a PRODUCT
whose feature set has `autoIntegrate: ENABLED` and one dependency component with
releases, so the mutation genuinely has work to do.

### 1. The bug reproduces on the shipped UI

Sandbox running the released `rearm-ui:26.09.20` (no fix). Playwright opens the
feature-set settings modal and issues ONE double-click on **Trigger Auto
Integrate**, counting the `autoIntegrateFeatureSet` POSTs the browser actually
sends:

```
[unfixed]   autoIntegrateFeatureSet mutations SENT: 2
[unfixed-2] autoIntegrateFeatureSet mutations SENT: 2   (repeat run)
```

### 2. The fix holds

Same probe, same fixture, UI hot-swapped to this branch's build:

```
[fixed]     autoIntegrateFeatureSet mutations SENT: 1
```

### 3. The duplicate is real, and the guard is correctly described as partial

Two **concurrent** calls -- exactly what the unfixed UI emits -- fired at the API
with a fresh dependency release each time:

```
attempt 1: concurrent results=[0.3.0, null]   product releases 3 -> 5
attempt 2: concurrent results=[0.3.3, 0.3.2]  product releases 5 -> 8
attempt 3: concurrent results=[0.3.5, 0.3.6]  product releases 8 -> 11
```

Attempts 2 and 3 each returned **two distinct product release versions from one
dependency change**. So the check-then-act race is not theoretical, and this
also confirms the narrowed comment: the client guard stops the double-click, it
does **not** close the race -- two API callers still duplicate, as above.

### 4. The review fixes, exercised against the paths they were written for

Against the fixed build, with the request layer manipulated to force each case:

**A -- the mutation fails** (fulfilled with a GraphQL error):

```
A: toast            -> Simulated failure for PR #277 verification | Auto Integrate Failed
A: reported failure -> true      <- the missing catch, now reporting
A: form edit kept   -> true      <- unsaved settings edits survive (value still "EDIT-KEEP-47664")
A: button usable    -> true      <- in-flight flag reset, title back to "Trigger Auto Integrate"
A: second click sent another mutation -> true
```

**B -- the mutation SUCCEEDS but every follow-up fetch fails** (the defect the
re-review caught: it used to report failure after a release had been created):

```
B: mutation was sent      -> true
B: toast                  -> Auto-integrate was attempted... | Auto Integrate Completed
B: WRONGLY reported fail  -> false    <- must be false, and is
B: reported an outcome    -> true
```

### Notes

- One test-harness trap worth recording: naive-ui notifications auto-dismiss at
  `duration: 3500`, so reading toasts once after a 4s wait finds nothing and
  **every toast assertion passes vacuously**. My first run reported "no toast"
  on all paths, including ones that definitely show one. The probe now polls and
  accumulates.
- Sandbox closeout done: image restored to `26.09.20`, `kubectl-set`
  field-manager stripped (`helm`, `kubectl-rollout`, `k3s` remain), node-local
  dev images pruned, cluster lane released.
- The fixture (`ai-prod-*` / `ai-dep-*` and its auto-integrated releases) is
  still in the agent sandbox org if anyone wants to re-run this; happy to tear
  it down.

logicflakes:
## Folded in the third outcome, verified both ways (`1d87df87`)

The flagged pre-existing path is now fixed here, since a reviewer would
otherwise have to hold a fixed error path and an unfixed success path in mind at
once.

**The problem.** The mutation has three outcomes, not two. It can fail; it can
succeed and create a release; or it can **succeed and create nothing** -- usually
because a matching product release already exists. Only the middle one closes
the modal. Both success outcomes were running `onCreated()`, which reassigns
`modifiedBranch` from `branchData`, i.e. the settings form. So on the third
outcome the operator's unsaved edits were reverted while they were still sitting
in the open form, with none of the confirmation that closing that same modal on
unsaved edits gives them.

**Verified before fixing, on the shipped `26.09.20` image** -- so it is
pre-existing, not something this branch introduced:

```
typed "UNSAVED-EDIT-55365" into the settings form
mutation result   : null (nothing created)
modal still open  : true
form value now    : "Base Feature Set"
EDIT SILENTLY LOST: true
```

**And after, same probe, same null result:**

```
form value now    : "UNSAVED-EDIT-17397"
EDIT SILENTLY LOST: false
```

**The fix** is the split the error path already makes: only the branch that
closes the modal earns a full `onCreated()`; the branch that leaves it open
refreshes the **releases** and touches nothing else. Not simply dropping the
call -- the list genuinely can be stale there, because a null result often means
something else created that release moments ago.

Three outcomes, three behaviours, one rule: never reset a form the operator is
still looking at.

**Regression check** -- re-ran the other two sandbox probes against this same
build:

```
[fixed-v2] autoIntegrateFeatureSet mutations SENT: 1
A: reported failure -> true | form edit kept -> true | button usable -> true
B: WRONGLY reported fail -> false | reported an outcome -> true
```

Sandbox closed out again: image back to `26.09.20`, `kubectl-set` field-manager
stripped (`helm`, `kubectl-rollout`, `k3s` remain), dev images pruned, cluster
lane released.

PR #279, opened 2026-08-11
refactor(ui): drop the team editor from user groups (Phase 2b)

logicflakes:
**Stacked on #278**, pairs with rearm-saas#411.

The **Member Roles** and **Notification Channels** sections come out of the user-group modal. A user group is a permission set again; who to contact and where a team is reachable belong to **Org Settings → Teams**.

Everything that existed only to serve them goes with it:

- the `teamRoles` / `teamChannels` editor state and its role setters;
- the pre-submit validation banner and `memberLabelFor` / `rosterMembers`;
- the channel picker's options builder and the org-channel loader;
- the entire separate team-fields query, with its `teamFieldsSupported` / `groupTeamMembers` / `teamMembersKnown` drift apparatus — that query had already degenerated to selecting just `uuid`.

`utils/teamMembers.ts` and its spec are deleted (the external half went in #274), as is the `TeamRoles` constant, since the backend enum it mirrored is gone.

## The payload builder collapses, but stays extracted

`buildUserGroupUpdateInput`'s omit-or-you-delete rules existed for the team fields. With those gone it's a plain projection — but it stays in `utils` rather than being inlined back, because the same REPLACE semantics still govern `manualUsers`, `connectedSsoGroups` and `permissions`, and a testable seam on the payload is what caught the last regression here.

Its spec keeps one assertion that no team key is ever sent. Belt and braces next to a schema that now rejects them — but the payload is where a reintroduction would start, and a group save silently clearing a team's channels is exactly what that file was extracted to prevent.

UI suite **143 → 132**, build clean, `validate-graphql` 0 Pro failures.


logicflakes:
Validated on the sandbox alongside rearm-saas#411 — the user-group modal no longer offers Member Roles or Notification Channels, groups still load and save with the legacy keys sitting in their `record_data`, and the Teams tab is unaffected. Backend probe 8/8; UI suite 132/132, build and `validate-graphql` clean.

PR #281, opened 2026-08-12
fix(ui): restore userGroupPermissionsDirty, and move assignment rules to Teams

logicflakes:
Two findings from rhythm walking the sandbox.

## 1. The edit-group modal could not be closed

Phase 2b (#279) deleted `userGroupPermissionsDirty` along with the block above it — a regex removing `teamMembersError` ran past its closing line. The modal's close handler reads it, so closing threw `ReferenceError` and the modal became a trap.

**Nothing caught it, and that is the part worth fixing in process:**

- `npm run build` compiles the SFC, but a template/handler reference to a missing setup binding only fails at **runtime**;
- all 132 unit tests cover `utils/` — there are no component tests;
- no browser probe had ever opened this modal.

Build-green plus tests-green said nothing at all about it. This is the third time this week that combination has proved hollow for a UI surface, after `npm run dev` being dead on main and the interpolated GraphQL documents opting themselves out of validation.

## 2. Team Assignment Rules were on the wrong tab, with the wrong data

They sat under **User Groups** and were handed `:teams="userGroups"`. Since Phase 2a a rule's `ownerTeam` is a **Team** — so that prop had been wrong since the switch, and the rule editor was offering the wrong entity entirely. rhythm spotted the placement; the broken prop came out of looking at why.

Moved into `TeamsOfOrg`, which already owns the real team list, with the parent passing `components` down for the match preview.

## Guard

`probe-usergroup-modal.js` covers both, and watches the **console**, not just the DOM — a modal that closes while throwing is still broken.

It separates auth-redirect noise from app-phase errors rather than filtering errors wholesale: the login page's own Keycloak theme serves stylesheets with a bad MIME type, and folding that in would make the assertion permanently red and therefore ignored.

**9/9** against the sandbox. 132/132 unit tests, build clean.

### DIFF
--- a/ui/src/components/OrgSettings.vue
+++ b/ui/src/components/OrgSettings.vue
@@ -611,18 +611,14 @@
                         </div>
                     </n-modal>
 
-                    <OrgTeamAssignmentRules
-                        :orgUuid="orgResolved"
-                        :isWritable="isWritable"
-                        :teams="userGroups"
-                        :components="allComponents"/>
                 </div>
             </n-tab-pane>
 
             <n-tab-pane v-if="isOrgAdmin" name="teams" tab="Teams">
                 <TeamsOfOrg
                     :orguuid="orgResolved"
-                    :isWritable="isWritable" />
+                    :isWritable="isWritable"
+                    :components="allComponents" />
             </n-tab-pane>
 
             <n-tab-pane name="programmaticAccess" tab="Programmatic Access" v-if="isOrgAdmin">
@@ -1162,7 +1158,6 @@ import CreateApprovalEntry from './CreateApprovalEntry.vue'
 import ScopedPermissions from './ScopedPermissions.vue'
 import OrgIntegrations from './OrgIntegrations.vue'
 import OrgGlobalApprovalPolicyRules from './OrgGlobalApprovalPolicyRules.vue'
-import OrgTeamAssignmentRules from './OrgTeamAssignmentRules.vue'
 import TeamsOfOrg from './TeamsOfOrg.vue'
 import AiAgentPoliciesOfOrg from './AiAgentPoliciesOfOrg.vue'
 import CommittersOfOrg from './CommittersOfOrg.vue'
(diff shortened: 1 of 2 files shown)


logicflakes:
## Both review layers run (`331bf8c7`) — I had skipped them, and they found a blocking bug

To be clear about what had and hadn't happened: this PR was built, deployed and probed 9/9, with 132/132 unit tests. **Neither review layer had run.** rhythm asked; that's the honest answer, and running them changed the PR.

### The blocking one, found independently by both layers

Moving the assignment rules to the Teams tab left their **data** behind. `loadTabSpecificData`'s `userGroups` branch dispatches `fetchProducts` under this comment:

> *Products as well as components: the team-assignment preview filters on type, so a PRODUCT/ANY rule would otherwise preview "matches 0".*

That comment exists **only because the rules lived there**. There was no `teams` branch at all, and `onMounted` fetches components but never products. So opening Org Settings → Teams directly would render *"Matches no components right now"* for every PRODUCT or ANY rule — indistinguishable from a genuinely bad pattern. Precisely the failure that comment was written to prevent, relocated one tab over.

No probe would have caught it: "0 matches" looks like a legitimate answer.

### Also fixed

- **`components` was optional with a `|| []` fallback** — which degrades silently into the same misleading "0 matches". One call site, always has the data; the prop is now required.
- **Three dead imports** left by the 2b cleanup. Nothing flags them: the eslint config enables no unused-symbol rules.
- **Four orphaned comments** from that cleanup — one had come to sit directly on top of the restored computed, describing a channel picker.
- A `Null until probed` comment over a ref that is never null, and a prop binding inconsistent with every sibling in its template.

### Verified rather than assumed

The restored `userGroupPermissionsDirty` is **character-identical** to what `ab3a39bb` deleted. And rather than hoping that regex ate only one thing, I extracted every declaration the 2b commit removed and checked each against the current file: exactly one was still referenced. Layer 2 reached the same conclusion independently across all 18 removed bindings.

### Systemic follow-up filed, deliberately not done here

`vue/no-undef-properties` catches this exact class. I trialled it across all of `src`: **one violation, and it's a false positive** (naive-ui scoped-slot props in `OrgIntegrations.vue`), so adoption costs two disable comments.

I did **not** enable it, because it would buy nothing today — `npx eslint src` already reports 53 pre-existing errors and no CI job runs lint or the UI tests at all. Adding a 54th failing rule to a lint run nobody gates on is theatre. Filed as `t20260812-073900-1182` with the right order: fix the 53, wire lint into CI, then turn it on.

Re-probed after all the above: **9/9**, deployed to the sandbox.

PR #282, opened 2026-08-12
feat(ui): hide durability, and tidy the subscription editor

logicflakes:
Three commits that were meant to be part of #281 but landed on the branch
after it merged. #281 was merged at 09:06Z carrying only its first two
commits; these three were pushed at 10:12, 10:38 and 11:20 to a branch whose
PR was already closed, so GitHub was never going to pick them up. Verified by
pointing the sandbox at the released `rearm-ui:26.08.19` (the #281 merge
build) and re-running the probe: 7 of 12 checks failed there -- PREVIEW back
in the status list, the rate-limit control back, `Notify Component Owner`
gone. Cherry-picked onto main; the resulting tree is byte-identical to the
branch head that was validated on the sandbox (`git diff b6501591 HEAD` is
empty).

## What is in here

**1. Stop surfacing durability (74301743).** The component Owner block showed
an ownership status chip -- OWNED / NON_DURABLE / DEGRADED / ORPHANED -- plus
a reason line explaining why a team was not durable. rhythm asked for the
label to go, then clarified: hide the distinction entirely, not collapse it
to "Owned". Both the word and the colour are gone, since a colour difference
leaks the distinction just as well as the word did. The owner NAME still
renders.

Also in this commit: when a subscription test produces no delivery, the
diagnosis now names TEAM targets. rhythm archived a targeted team and was
told "its filter, or a route's minimum-severity gate, likely excluded it" --
both innocent. The team cause now leads and the filter/severity guesses are
demoted.

**2. Remove the ownership status chip entirely (fd0f25f4).** The follow-up to
the clarification above.

**3. Tidy the subscription editor (b6501591).** Six items from rhythm's
walkthrough:

- PREVIEW removed from the status select, along with its apology paragraph.
  PREVIEW is silently identical to DISABLED (t20260811-030648-23714); a spec
  is filed as t20260717-120722-18192 if it is ever wanted.
- Filter mode already defaulted to Preset -- verified rather than "fixed".
  The subscription rhythm saw in ADVANCED is genuinely stored that way.
- Minimum severity is shown only for the two event types that carry a
  severity. This is the one worth reading closely: `severityGateMatches`
  treats "gate set, severity null" as NO MATCH, so a minimum severity on a
  release-or-approval subscription does not do nothing -- it silently gates
  out every event. Hiding the control alone would have preserved the trap
  somewhere unreachable, so the stored value is cleared when the control
  hides.
- Teams and Notify Component Owner descriptions moved to label tooltips
  (`InfoCircle`, matching the house pattern).
- The rate-limit control is hidden. `rateLimit` is parsed, stored and echoed
  back, and never read at fan-out (t20260702-065833-25754, decided
  leave-as-is). The form still loads and re-sends the stored value, so an
  existing limit round-trips rather than being silently cleared by an
  unrelated edit, and the control returns when enforcement lands.

One thing deliberately NOT moved to a tooltip: the conditional caveat under
the owner checkbox ("this org has no teams yet, so this delivers nothing").
That is not a description, it is a statement that ticking the box achieves
nothing right now. A caveat behind a hover is a caveat nobody reads.

## Validation

- `npm run test`: 136 passed (12 files). Nothing in CI runs this suite
  (t20260811-061400-10264), so it is part of manual validation.
- `npm run build` green; `validate-graphql.mjs` 0 Pro failures, warnings are
  the expected CE-mirror lag on Team (backend is rearm-saas only).
- Sandbox: `probe-subscription-editor-tidy.js` 12/12, and
  `probe-durability-hidden-and-team-hint.js` for the chip removal, driven
  through a team that the backend still computes NON_DURABLE so the check
  cannot pass vacuously.

## Reviews

Layer 1 and Layer 2 ran on #281 at `331bf8c7` -- which predates all three of
these commits. Both layers are being re-run against this diff.

### DIFF
--- a/ui/src/components/ComponentView.vue
+++ b/ui/src/components/ComponentView.vue
@@ -1105,6 +1105,9 @@ import graphqlClient from '../utils/graphql'
 import constants from '@/utils/constants'
 import { InputTriggerEvent } from '../utils/triggerTypes'
 import { validateInputTrigger, validateOutputTrigger } from '../utils/triggerValidation'
+// Shared "keep a dangling reference visible and removable" builder, so the owner
+// picker behaves like the notification channel and team pickers.
+import { withGhosts } from '@/utils/channelOptions'
 import CelExpressionBuilder from './CelExpressionBuilder.vue'
 import graphqlQueries from '../utils/graphqlQueries'
 
@@ -1501,10 +1504,15 @@ const openComponentSettings = async function() {
     // Durable ownership (Phase 4 UI): load the owner picker's team list + the
     // computed ownership status when the settings panel opens.
     //
-    // AWAIT the team list FIRST. loadOwnership pre-selects the stored owner only
-    // if it resolves to an option, so racing these would leave a perfectly valid
-    // owner unselected whenever the ownership query happened to win.
-    await loadUserGroups()
+    // AWAIT both option lists FIRST. loadOwnership pre-selects the stored owner
+    // only if it resolves to an option, so racing these would leave a perfectly
+    // valid owner unselected whenever the ownership query happened to win. Users
+    // are in here too: they are otherwise loaded fire-and-forget from initLoad,
+    // which is the same race one field over for a USER-typed owner.
+    // loadUserGroups swallows its own errors; loadUsers does not, and a rejection
+    // here would abort openComponentSettings before it ever shows the modal. A
+    // failed lookup must cost one empty picker, not the whole settings panel.
+    await Promise.all([loadUserGroups(), loadUsers().catch(() => {})])
     await loadOwnership()
     originalComponent.value = commonFunctions.deepCopy(updatedComponent.value)
     showComponentSettingsModal.value = true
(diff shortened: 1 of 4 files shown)


logicflakes:
## Two-layer review: run, findings fixed

Both layers ran against the settled diff (Layer 1 `/review`, Layer 2 a
ReARM-conventions subagent). Fixes are in `f875c346`.

**Both layers independently found the same blocking bug, and it was one this PR
introduced.** The severity gate was hidden but not cleared on the common path:
the clearing is a `watch()` on a computed, so opening a subscription that is
ALREADY release-only-with-a-gate never fires it, and the save re-persists the
value. That is strictly worse than before the control was hidden -- the gate
still suppresses every event, and now nobody can see or reach it.

I verified it live before fixing rather than reasoning from source, because the
whole point of the finding is that the code reads as if it works:

```
stored gate before=HIGH   after an edit+save=HIGH     <- before the fix
stored gate before=HIGH   after an edit+save=null     <- after
```

Layer 2 additionally caught that an archived owner team now rendered as *no
owner at all* (name gone, picker empty, bare "Clear owner" button), and that
this repo already has a shared answer to exactly that -- `withGhosts`, which the
subscription team picker uses for the identical case. The owner picker now uses
it too, so the two cannot drift.

Full list in the commit message. Other items: `users` awaited alongside teams
(the USER-owner branch was still racing), `rateLimit` raw-blob stash so a
partial limit is not silently deleted by an unrelated edit, `noDeliveryExplanation`
moved into the util and given tests, several comments that had stopped being
true, and tooltips brought onto the house pattern.

**Validation:** 136 -> 147 unit tests (the multi-route clearing test was
mutation-checked -- clearing only `routes[0]` fails it), build green,
`validate-graphql` 0 Pro failures, added lines plain ASCII. On the sandbox:
stale-gate 3/3, archived-owner 5/5, editor tidy 12/12, durability + team-hint
9/9.

**Deliberately not fixed here:** the perspectives gate has the same trap shape
as severity (`perspectiveGateMatches` returns false when the event carries no
`affectedReleases`) but is still shown for every event type. Hiding it rests on
a premise I have not verified the way I verified the severity one, and hiding a
control on an unverified premise removes a feature people may be using. Noted in
the template and filed on the board.

PR #283, opened 2026-08-12
docs: document Teams, and correct what the Team work changed

logicflakes:
The notification and ownership pages were written before Teams became a
first-class entity. They referenced teams as something the reader was assumed
to already know, pointed at a tab that no longer hosts assignment rules, and
described durability by a rule that has since changed.

## New page: Teams

`configure/teams.md` -- what a team is (accountability, not access), what it
holds, why you would contain a user group instead of listing people, and what
archiving one actually does. That last part is the one worth writing down:
archiving a team silently stops delivery on every route targeting it and
degrades ownership of every component it owns, and nothing fails anywhere while
that happens.

## Corrected, each checked against the code rather than carried over

- Assignment rules live under **Organization Settings -> Teams** now, not User
  Groups (they moved with #281).
- Durability is measured against the team's **roster** -- members plus everyone
  contributed by contained user groups -- or a contained SSO-backed group. The
  old text said "two direct members", which now understates a team of one that
  contains a ten-person group.
- The component Settings panel no longer reports a durability status, so the
  page no longer says it does; it points at the ownership report, which is where
  a fleet-level question belongs.
- The owner picker offers active teams only but still shows an already-stored
  archived owner, labelled. Documented because the alternative reading of an
  empty picker is "this component has no owner".

## Added -- existing behaviour, undocumented, easy to get wrong

- A minimum severity on a subscription whose event types carry no severity does
  not narrow it, it **silences it completely**. Verified against
  `extractEventSeverity` and `severityGateMatches`.
- A dedup window of 0 means "deliver everything", and dedup **cannot** be
  demonstrated with the Test button, because synthetic events skip it by design.
  Testing twice always produces two deliveries, which reads exactly like dedup
  being broken.
- A test that reports no delivery has a third explanation beyond filter and gate
  whenever a route resolves through a team or an owner.

## One claim I set out to document and found to be false

The UI hint and the `perspectiveGateMatches` javadoc both say perspectives make
release and approval events "match nothing", because only vulnerability events
carry `affectedReleases`. They do carry them: `ReleaseChangeHookImpl` (created /
lifecycle / bom-diff) and `ApprovalEventNotifierImpl` (requested / resolved)
both stamp `buildAffectedReleases`, and that helper copies the component's
perspectives onto each entry. A perspective-scoped route matches those event
types fine -- what it needs is a component that belongs to the perspective.

The page now says that. The stale UI hint is fixed in #282; the identical
javadoc claim lives in rearm-saas and is on the board
(`t20260812-121545-31450`).

## Validation

`npm run docs:build` clean, with VitePress dead-link checking on (no
`ignoreDeadLinks`), so every new cross-reference and anchor resolves. Added
lines are plain ASCII.

### DIFF
--- a/documentation_site/docs/.vitepress/config.mjs
+++ b/documentation_site/docs/.vitepress/config.mjs
@@ -32,6 +32,7 @@ function sidebar() {
         ]},
         {text: 'Configure', link: '/configure/', items: [
           {text: 'Users and User Groups Permissions', link: '/configure/user-and-user-group-permissions'},
+          {text: 'Teams', link: '/configure/teams'},
           {text: 'Component Ownership', link: '/configure/component-ownership'},
           {text: 'Notifications', link: '/configure/notifications'},
           {text: 'Approval Queues', link: '/configure/approval-queues'}
--- a/documentation_site/docs/configure/component-ownership.md
+++ b/documentation_site/docs/configure/component-ownership.md
@@ -27,9 +27,10 @@ person leaves, and only a team can be a notification target -- a user has no
 channels of their own, so a route that notifies the component owner will not
 deliver for a user-owned component.
 
-The picker offers active teams only. If the component is already owned by a team
-that has since been archived, that team is still shown -- labelled as archived --
-so the stale owner is visible and replaceable rather than an empty field.
+The picker offers active teams only, so an archived team cannot be chosen as a
+new owner. A component already owned by a team that was archived since keeps
+that owner on record -- it reports `DEGRADED` and stops receiving owner-routed
+notifications until you point it somewhere else.
 
 ## Ownership status
 
@@ -107,11 +108,11 @@ When more than one source could decide an owner, ReARM resolves in this order:
 
 ## Finding components at risk
 
-On an individual component, the **Settings** panel names the current owner. It
-does not report durability: that is a fleet-level question -- "which components
-have nobody durable behind them" -- and answering it one component at a time
-invites reading a single `NON_DURABLE` as a fault to fix rather than as coverage
-to plan.
+On an individual component, the **Settings** panel names the current owner.
+
+Durability, though, is a fleet-level question -- "which components have nobody
+durable behind them" -- and answering it one component at a time invites reading
+a single `NON_DURABLE` as a fault to fix rather than as coverage to plan.
 
 Organization-wide, an **ownership report** returns every component that is not
 cleanly `OWNED`, together with its computed status. This is the practical
--- a/documentation_site/docs/configure/component-ownership.md
+++ b/documentation_site/docs/configure/component-ownership.md
@@ -14,17 +14,22 @@ have nobody durable behind them.
 Ownership is deliberately **separate from access**. Owning a component does not
 grant permissions on it, and holding permissions does not make you the owner.
 Access is configured under
-[Users and User Groups Permissions](./user-and-user-group-permissions).
+[Users and User Groups Permissions](./user-and-user-group-permissions);
+accountability is carried by a [team](./teams).
 
 ## Setting an owner
 
 Open a component, go to **Settings**, and use the **Owner** picker. You can
-name a team or an individual user.
+name a [team](./teams) or an individual user.
 
 Prefer a **team**. An individual owner is valid and ReARM records it, but a
 person leaves, and only a team can be a notification target -- a user has no
 channels of their own, so a route that notifies the component owner will not
 deliver for a user-owned component.
+
+The picker offers active teams only. If the component is already owned by a team
+that has since been archived, that team is still shown -- labelled as archived --
+so the stale owner is visible and replaceable rather than an empty field.
 
 ## Ownership status
 
@@ -34,7 +39,7 @@ it always reflects the current state of the owner it points at.
 | Status | Meaning |
 |---|---|
 | `OWNED` | A stored owner that is valid and durable. The healthy state. |
-| `NON_DURABLE` | Valid, but resting on a single point of failure: an individual owner, or a team with fewer than two direct members that is not SSO-backed. |
+| `NON_DURABLE` | Valid, but resting on a single point of failure: an individual owner, or a team whose roster is smaller than two people and that contains no SSO-backed group. |
 | `DEGRADED` | The owner still resolves but has weakened -- typically the owner team has been archived or made inactive. |
 | `UNSET` | No owner has been chosen, but ReARM can suggest a candidate team. |
 | `ORPHANED` | No owner and nothing to suggest, or a stored owner that no longer resolves. Needs a human. |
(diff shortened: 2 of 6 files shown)


logicflakes:
## Two-layer review: run, findings fixed

Layer 2 (ReARM-conventions subagent) and Layer 1 both ran. Fixes are in
`89da155e` and `d2251792`.

**Layer 2 independently confirmed the perspectives reversal** against the
producers -- `ReleaseChangeHookImpl` for all three release events,
`ApprovalEventNotifierImpl` for both approval events, and
`ReleaseNotificationSupport.buildAffectedReleases` stamping the component's
perspectives onto each entry. It also checked every other claim on the page
against the named source: org-admin gating, the field table, roster semantics,
lead validation, name uniqueness, archiving consequences, durability, the
severity gate, dedup-0 and the synthetic bypass. Those stand.

**The real finding was merge order.** Three sentences described UI that exists
only on #282, not on `main`:

- "the Settings panel does not report durability" -- on `main` it still shows the
  status chip and reason line;
- "an archived owner stays visible, labelled" -- on `main` the owner picker drops
  archived teams and renders a raw uuid;
- the `(archived)` label itself -- the route-target picker said `(deactivated)`.

All three are now written so they are true on `main` today and remain true after
#282, rather than betting on merge order. Where a fact genuinely could not be
stated both ways, the page says the thing that matters instead: an archived
owner stays on record, reports `DEGRADED`, and stops receiving owner-routed
notifications.

Layer 2 also caught four text-level errors worth naming, because each was me
writing plausible prose instead of reading the code: the name-collision advice
told readers to *infer* an archived team when `requireNameAvailable` names it
outright; "set a team to inactive" named no control (it is **Archive**); the
rationale for name-and-description-only creation was invented, where
`CreateTeamDto`'s javadoc gives the real one; and "every event type" overreached
past `VEX_STATE_CHANGED`, which has no producer.

Layer 1 found two more of the same kind: a CE note that described CE's ownership
model in a clause that could not be right (the mirror has
`ComponentOwnershipService` but no `Team`), and the last surviving
forward-looking label claim.

**Out of scope, filed:** the `perspectiveGateMatches` javadoc in `rearm-saas`
still carries the false claim this page corrects (`t20260812-121545-31450`). The
UI half is fixed in #282, along with normalising the route picker's team label to
`(archived)` so one state has one word across all three pickers.

**Validation:** `npm run docs:build` clean with VitePress dead-link checking on
(no `ignoreDeadLinks`), so every cross-reference and anchor resolves. Added lines
are plain ASCII.

logicflakes:
## Refreshed against main (2026-09-18)

The branch was cut from #281 on 12 Aug and sat for five weeks. Merged
`origin/main` in (**not** rebased -- a rebase needs a force-push, which is
barred on this repo and would strip commit signatures) and did a targeted pass
on what changed underneath it.

**Merge:** one conflict, in the sidebar -- this branch and #356 each added an
entry at the same position under *Configure*. Kept both.

**What had drifted, and what the pass did about it**

| Landed after the branch | Fix |
|---|---|
| #285 team notifications for owned components | New `teams.md` section + a row in *What a team holds*. Covers the opt-out storage (a new event type arrives **selected**), the two warning cases, and the fact that it stacks with owner routing rather than replacing it -- dedup is per subscription, so an overlapping pair double-delivers. Also documents the managed subscription being visible-but-uneditable in the list, and that archiving a team **disables** that row rather than deleting it. |
| #287 owner assignment joins the staged save flow | *Setting an owner* no longer describes a self-committing picker. It stages behind the shared Save Changes / Reset Changes, and a cleared picker stages a *clear* -- which hands the component back to the org's assignment rules, not to nothing. |
| #282 / #284 | Two hedges written while these were unmerged are now unnecessary. An archived team that already owns a component is again documented as shown, labelled `(archived)`. Added the adjacent case: an owner that resolves to nothing leaves the picker **empty** rather than printing a bare uuid. The perspectives correction this PR already carried matches the corrected hint in #284 -- no change needed there. |

**Deliberately not done:** instance-event subscriptions (#293) are missing from
the event-type table on `notifications.md`, but that gap is on main and predates
this branch -- documenting it here would widen a docs-sync PR into an unrelated
one. Worth its own task.

**Validation:** docs site builds clean with dead-link checking on; the new
anchors were checked in the rendered HTML rather than assumed. Every claim added
above was read off the source it describes -- `TeamsOfOrg.vue`,
`teamNotificationEventTypes.ts`, `ComponentView.vue`'s owner block, and
`TeamService.syncOwnedComponentSubscription` for the archive/disable semantics --
not carried over from the PR descriptions.

logicflakes:
## Layer 2 (conventions + accuracy) run, findings fixed in `ae1ba279`

Ran the skeptical-senior conventions pass over the refreshed diff. Six findings,
four factual -- all the same failure mode this PR exists to fix, prose written
against August's code. Each was re-verified against the source before fixing,
which is worth noting because one finding was the reviewer catching a claim I
had sourced from a **code comment** instead of the code.

| # | Finding | Fix |
|---|---|---|
| 1 | **Teams are not Pro-only any more.** The stack landed in the CE backend with the shared-code sync two weeks after this branch was cut, with no edition gate; the tab gates on org-admin, and the "not available on this server version" alert keys off schema drift, not edition. | The tip now says what the alert says -- teams need a server build that has them -- instead of naming an edition. That claim also cannot go stale on release timing. |
| 2 | Same root cause in the route-target table and the edition aside. | **Teams** row is CE and Pro. **Notify the component owner** stays Pro, but described as the UI withholding it, since the CE schema does carry the field. |
| 3 | "Only two event types carry a severity" -- true when written, not now: `extractEventSeverity` gained both instance-deployment types, which compute one. | Four, with the carry-none list spelled out by name. The warning's actual point is unchanged and still checks out against `severityGateMatches`. |
| 4 | **The unresolvable-owner sentence I added was wrong.** An unresolvable ref seeds a null baseline, and `stagedOwnerChange` returns `'none'` without one -- so clear is exactly what you cannot stage there. It also conflated two cases. | A team ref that resolves to nothing keeps a `(deleted team) <short id>` ghost and clears fine; only an individual owner leaves the picker empty, and from there you must pick rather than clear. |
| 5 | Managed-row controls are Edit / Enable-Disable / Delete, not "pause"; Test and delivery history stay live. | Reworded, and the still-live actions named -- it is why the row is visible at all. |
| 6 | CE-alert wording scoped to the wrong thing. | Folded into #1. |

**Explicitly clean** (the reviewer verified, no change needed): the perspectives
paragraph -- correct despite `perspectiveGateMatches`' javadoc still claiming
the opposite, which is a stale comment worth its own cleanup; dedup window `0`;
the test-dialog troubleshooting order; the opt-out storage rule and both team
warnings; the archive DISABLE-not-delete semantics; name uniqueness including
archived names; the durability rewrite; and all 35 internal links/anchors.

**Validation:** docs build clean with dead-link checking on; ASCII byte sweep
clean for added text (its two hits are pre-existing lines in files this branch
touches).

Not done, still deliberate: instance-event subscriptions (#293) remain absent
from the event-type table -- a gap on main, not this branch's to close.

PR #286, opened 2026-08-12
fix(ui): stop remounting the whole view on query-only navigation

taleodor-claude:
## Problem

Switching Integrations sub-tabs (Catalog / Subscriptions / Channel groups) flickers the Catalog "N active" count (e.g. 5-6-7 back and forth). Same class of jank exists on every URL-synced tab/filter in the app.

## Root cause

`LeftNavBar`'s router-view is keyed by `$route.fullPath`, so **every query-only navigation destroys and remounts the entire route component**. Each Integrations sub-tab click remounted OrgSettings → OrgIntegrations, refiring the four catalog fetches (configured integrations, CI integrations, notification channels cached; BEAR always network-only) — the count climbs through their staggered arrivals on every click.

## Fix

- Key the router-view by `$route.path`. Entity identity lives in path params on every route, so entity→entity navigation still remounts; query changes leave the view alive.
- The five views that sync UI state to query params now apply query changes via watchers — this covers browser back/forward and cross-component query pushes, which previously worked only *because* of the remount:
  - `OrgIntegrations` (`integrationsTab`), `InstanceView` (`instTab` + lazy history fetch), `FindingAnalysisPage` (`tab` — its `onTabChange` wrote only the URL and depended on the remount entirely), `MostRecentReleasesPage` (date/limit/type), `VulnerabilityAnalysis` (filter chain + `cveId` modal deep-link, per-ref guards so identical re-pushes can't loop).
  - `OrgSettings` already watched its `tab`/`auditTab`/`policyTab` params — no change needed.

## Verification

- vue-tsc clean (pre-existing tsconfig deprecation notices aside), vitest 147/147, vite build green.
- psclaude smoke after branch image build.

### DIFF
--- a/ui/src/components/BranchView.vue
+++ b/ui/src/components/BranchView.vue
@@ -358,7 +358,7 @@ export default {
 }
 </script>
 <script lang="ts" setup>
-import { ComputedRef, computed, Ref, ref, h, watch } from 'vue'
+import { ComputedRef, computed, Ref, ref, h, watch, nextTick } from 'vue'
 import { useStore } from 'vuex'
 import { useRoute, useRouter, RouterLink } from 'vue-router'
 import { NButton, NCheckbox, NForm, NFormItem, NInput, NModal, NPagination, NPopover, NSelect, NotificationType, useNotification, SelectOption, NDataTable, NIcon, NSpace, NSpin, NTag, NTooltip, DataTableColumns, NSelect as NSelectComponent, NDropdown} from 'naive-ui'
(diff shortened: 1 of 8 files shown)


taleodor-claude:
Review addressed in 08568398 — all three findings were valid:

**Finding 1 (leave-fire push junk):** exactly as described. Fixed with the suggested shape: route-name guard on the URL watcher plus an `applyingFromUrl` flag that the selector watchers consult before running their clear-downstream + push cascade. URL-originated writes now refresh the dependent option lists explicitly, and the records-refetch watcher stays ungated (it only fetches). This also fixes the back/forward history-entry minting — restoring an entry no longer pushes a differing query. Live-verified with a new probe (`rearm-playwright/inspect-va-leave-junk.js`): navigating away from `/vulnerabilityAnalysis?componentProduct=…` leaves the destination URL clean, zero page errors.

**Finding 2 (ComponentView):** `tab` is now watcher-applied with absent = `branches` (mirroring the old remount initializer, and clearing the selected branch like `handleTabChange` does); the `componentSettingsView` deep link opens under a route-name guard (the async open cascade must not fire on leave) and closes on param removal with remount parity (no unsaved-changes prompt).

**Finding 3 (BranchView):** pagination gets its own applying-flag (restoring a perPage entry would otherwise push page-1 and destroy the forward stack); `?release=` is watcher-applied — note local open/close writes the URL via `history.replaceState`, which vue-router never sees, so the watcher fires exclusively for real navigations.

Minor notes: agreed on both — MostRecentReleasesPage left as-is (replace-based sync, pristine entries get replaced; inconsistency noted) and PullRequestsOfOrg untouched.

vue-tsc clean, vitest 147/147, vite build green; psclaude runs `2026-08-query-nav-no-remount.2` with the flicker + leave-junk probes both PASS.

PR #287, opened 2026-08-12
feat(ui): owner assignment joins the staged component-settings save flow

taleodor-claude:
## Problem

In component settings, the Owner (team assignment) picker was the one control wired differently from everything else: dedicated **Set owner** / **Clear owner** buttons committed immediately, while every other core setting stages into `updatedComponent` behind the shared **Save Changes** / **Reset Changes** pair.

## Change

Owner edits now stage like the rest:

- The picker (now `clearable`; buttons removed) writes only draft state. `stagedOwnerChange` derives `none`/`set`/`clear` against the baseline the picker was **seeded** with — deliberately not the raw stored owner, so a dangling ref (deleted user) that seeds an empty picker can't stage an accidental clear, and toggling the Team/User dropdown (which auto-empties the ref for browsing) stages nothing.
- The shared `save()` carries `owner` / `clearOwner` on the same `UpdateComponentInput` the dedicated mutation used, and re-reads ownership on success. The store `updateComponent` action's field whitelist now passes `owner`/`clearOwner` through — it silently dropped them before, which is why the old code bypassed the store with its own mutation (now deleted).
- `hasCoreSettingsChanges` surfaces the shared buttons for a staged owner edit; `resetCoreSettings` and the modal close-guard (unsaved-changes prompt) revert the draft.

## Verification

- vue-tsc clean, vitest 147/147, vite build green.
- psclaude smoke on the branch build: stage owner → shared buttons appear → Save persists → label + baseline refresh; Reset reverts; clear via picker X stages `clearOwner`.

### DIFF
--- a/ui/src/components/ComponentView.vue
+++ b/ui/src/components/ComponentView.vue
@@ -409,6 +409,10 @@
                                                      it. The backend still computes every status for the ownership
                                                      report, which is where coverage belongs. -->
                                                 <div v-if="ownerLabel">{{ ownerLabel }}</div>
+                                                <!-- Staged like every other core setting: picking (or clearing —
+                                                     the X hands the component back to org team-assignment rules)
+                                                     only stages the change; the shared Save Changes / Reset
+                                                     Changes buttons commit or revert it with the rest. -->
                                                 <div v-if="isWritable" style="display: flex; gap: 8px; margin-top: 8px; align-items: center;">
                                                     <n-select
                                                         style="width: 110px;"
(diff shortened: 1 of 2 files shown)


taleodor-claude:
Review addressed in ae83af2d — the finding was real and precisely diagnosed: the toggle-away guard protected the toggled-away state but not the round trip, since `@update:value` re-nulled the ref on the way BACK to the baseline type, landing on exactly the staged-clear shape. The type handler now re-seeds the ref from the baseline when returning to the baseline type and nulls it only when switching away, per the suggested fix.

Verified on psclaude (`2026-08-owner-staged-save.1`) with a new probe leg: User → Team → User re-seeds the picker and stages nothing (no shared buttons appear); the full probe is 7/7 including the intentional-clear leg, which still works.

On the non-blocking note: acknowledged — a dangling stored owner (deleted user) can no longer be cleared directly; the baseline seeding that prevents accidental clears is what makes the intentional one unreachable there. Given the rarity and the two-step workaround (set a real owner, save, clear), leaving it as documented behavior rather than special-casing — recording it here for discoverability.

PR #288, opened 2026-08-12
fix(rebom): built-in BEAR enrichment skip for unresolvable purl types

taleodor-claude:
## Problem (production)

An SBOM produced by a filesystem-cataloguing scanner carried thousands of per-file `pkg:generic` components (OS files recorded as `pkg:generic/<file>?path=...`). `pkg:generic` has no upstream registry by definition, so BEAR has nothing to resolve those purls against — but the enrichment run asked anyway, spent its entire 30-minute budget on them, timed out, and was retried from scratch by the scheduler every cycle: `FAILED -> retry -> FAILED`, indefinitely (there is no retry cap — the scheduler's own TODO).

The blast radius is org-wide, not one BOM: the backend enrichment puller takes the org's **oldest** un-enriched matchable components (400-wide window), so the stuck BOM's components monopolized the window, every tick resolved to the one PENDING/FAILED BOM, zero pulls happened, and every other BOM's components queued behind them un-stamped — org-wide `[SYNTHETIC-STALL]` / `[FANOUT-STALL]`.

## Fix

A built-in skip-pattern list (`BUILT_IN_ENRICHMENT_SKIP_PATTERNS`, currently `pkg:generic/`) **merged with — never replaced by — the org's configured `integration.config.skipPatterns`**, applied at the single point every enrichment run passes through (`enrichCycloneDxBom`). Org configuration can add patterns but cannot remove a built-in.

Why skipping is safe:
- Skipped components still ship: rearm-cli copies them into the enriched output untouched, the BOM reaches `COMPLETED`, and the backend puller stamps `enriched_at` on every component the parsed BOM carries — the synthetic Dependency-Track gate passes and the org unblocks.
- No signal is lost: no vulnerability source indexes `pkg:generic` (no OSS Index / OSV / GHSA ecosystem, no CPE), and license identity for OS files belongs to their *package* entry, not per-file rows.

The merge rule lives in `enrichmentSkipPatterns.ts` as a pure function with 6 tests, not inline in the service.

## Verification

- `tsc --noEmit` clean; unit suite vs live rebom PG **171/171** (165 pre-existing + 6 new).
- Follow-up (separate, Pro backend): mark `pkg:generic` components enrichment-terminal so they also stay out of synthetic Dependency-Track buckets and the coverage gate.

### DIFF
--- a/rebom-backend/src/services/bom/bomProcessingService.ts
+++ b/rebom-backend/src/services/bom/bomProcessingService.ts
@@ -18,6 +18,7 @@ import * as SpdxRepository from '../../spdxRepository';
 import { getBearCredentials, getBearIntegration } from '../integrationService';
 import { SpdxService } from '../spdx';
 import { downgradeCycloneDxSpecIfNeeded } from '../cyclonedx/cdxSpecDowngrade';
+import { effectiveSkipPatterns } from './enrichmentSkipPatterns';
 import { SPDX as CDXSpdx } from '@cyclonedx/cyclonedx-library';
 const canonicalize = require('canonicalize');
 import { createHash } from 'crypto';
@@ -639,13 +640,13 @@ export async function enrichCycloneDxBom(
       '-o', outputFile
     ];
     
-    // Add skip patterns
-    for (const pattern of skipPatterns) {
+    // Built-in skip patterns (unresolvable purl types) merged with the
+    // org-configured ones — see enrichmentSkipPatterns.ts for why.
+    const patterns = effectiveSkipPatterns(skipPatterns);
+    for (const pattern of patterns) {
       args.push('--skipPattern', pattern);
     }
-    if (skipPatterns.length > 0) {
-      logger.debug({ skipPatterns }, 'Added skip patterns to enrichment command');
-    }
+    logger.debug({ skipPatterns: patterns }, 'Added skip patterns to enrichment command');
 
     logger.info('Running enrichment: rearm-cli bomutils enrich');
     const result = await shellExec('rearm', args, ENRICHMENT_TIMEOUT_MS);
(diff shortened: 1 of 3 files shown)


taleodor-claude (commented):
Reviewed with claim verification against rearm-cli and rebom sources — the design is sound and the load-bearing claims are accurate:

- **Substring semantics confirmed**: `shouldSkipPurl` is case-insensitive `strings.Contains` (cmd/bear.go:380), so `pkg:generic/` matches the observed `pkg:generic/<file>?path=...` shape. Patterns ride `shellExec` array args — no shell-quoting hazard.
- **Copy-through confirmed**: the skip is a `continue` in the enrichment-candidate loop; skipped components remain in the output BOM untouched, so the COMPLETED/`enriched_at` unblock chain in the PR body holds.
- **Single choke point confirmed**: all three `enrichCycloneDxBom` call sites pass `credentials.skipPatterns` and the merge happens inside the function — nothing bypasses the built-ins.

One minor hardening suggestion, non-blocking: `effectiveSkipPatterns` validates `pattern.trim().length > 0` but pushes the **untrimmed** string. A configured pattern with accidental whitespace (`" pkg:swift/"`) then reaches the substring matcher and can never match a purl (purls percent-encode spaces) — a silently dead config entry. Pushing `pattern.trim()` also makes the dedupe catch whitespace variants.

Drive-by finding, out of scope for this PR but worth an issue on rearm-cli: `addMetadataComponentToSkipPatterns` appends `.*<name>.*` — regex syntax fed to the same `strings.Contains` matcher, which can only match a purl containing the literal characters `.*`. The BOM-root self-skip has therefore never actually skipped anything; the root component is being enriched despite the intent.

taleodor-claude (commented):
Re-checked 71394bfd: exactly the suggested shape — trim before dedupe and push, non-strings skipped, and the two new tests pin both the dead-entry case and the whitespace-padded-built-in dedupe. CI green. LGTM — formal approval not possible (same PR author identity), so recording it here.

PR #310, opened 2026-09-01
fix(ui): branchless dependencies render as exclusions and require IGNORED to save

taleodor-claude:
Companion to the backend rule that a dependency without a branch pin is only legal when its status is IGNORED.

## Changes (BranchView)

- **Branch column**: an IGNORED branchless row renders as `— (excluded)` with the existing excluded styling; a non-ignored branchless row (data predating the rule) shows `No branch — select one or exclude` instead of `Unknown`, telling the operator exactly how to fix it.
- **Save validation**: saving with a non-excluded dependency lacking a branch fails fast client-side, naming the component and the two ways out (select a branch / set Excluded), instead of a server-side rejection after the round-trip.

Requires the backend schema change making `EffectiveDependency.branch` nullable to be deployed first — until then the whole `effectiveDependencies` block nulls out at the GraphQL layer regardless of UI handling.

## Testing

`npm test` — 176 tests, only the pre-existing `routeInputSchemaDrift` failure that fails identically on main; `npm run build` clean.

### DIFF
--- a/ui/src/components/BranchView.vue
+++ b/ui/src/components/BranchView.vue
@@ -777,6 +777,19 @@ async function saveModifiedBranch () {
             )
         }
         
+        // A dependency without a branch is only legal when IGNORED (matches
+        // backend updateBranch validation) - fail fast with a pointed message
+        // instead of a server-side rejection.
+        const branchlessInvalid = (modifiedBranch.value.dependencies || []).find(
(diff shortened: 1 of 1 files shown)


taleodor-claude:
Closing — superseded: migration tooling now resolves all dependencies to base branches, so branchless dependencies never reach the UI and the non-null schema stays as-is.

PR #318, opened 2026-09-04
FDA 7e UI (1/4): support display scaffolding + server-side attestation filter and keyset paging

logicflakes:
First of four PRs for the 7e attestation UI. Pairs with `rearm-saas#484` (merged to `fda-readiness-1-dev` there).

**Not a CE schema sync.** The paged query is Pro-only for now, by design — syncing `schema.graphqls` into CE is the one-way door on this track and is its own PR. This PR degrades gracefully on CE instead; see below.

## What's here

1. **Restored the support display scaffolding** #311 removed — the tag/risk maps, badge helpers, Support column, and the query fields that feed them. Not restored: the edit modal, the single-component mutation, the client-side filter and the coverage counter. Each now has a *different* server API (bulk with per-item outcomes, server-side filter with keyset paging, release-scoped coverage), so restoring them would mean writing against an API that no longer exists. They're PRs 2–4.
2. **Server-side `attestation` filter + keyset paging**, replacing the client-side search.

## The two couplings fixed on the way in

Both were the UI claiming it can render values the server cannot send:

- `SUPPORT_TAG` keyed on all six `SupportStatus` members with a `|| SUPPORT_TAG.UNKNOWN` fallback. Three are `@Deprecated` and never returned by `derive()`. The fallback was the worse half: **`UNKNOWN` is itself a meaningful status** meaning "not assessed", so falling back to it renders a value we don't understand as a confident *"nobody has assessed this"* — on a regulatory-disclosure screen. Now a distinct "Unrecognised status" tag plus `console.error`. Loud, not fatal: throwing takes the table down over one row.
- `DEVICE_RISK_LABEL` still carried `EOL_BEFORE_DEVICE`, removed from the backend enum with the end-of-sale redefinition.

## Review: both layers run, 12 findings

The two that mattered were things this change *broke*:

**The CE fallback was aimed at a query CE cannot serve.** This UI ships in the CE repo and the CE mirror has neither the paged query nor the filter enum, so a CE install rendered the tab as a toolbar over nothing. Writing the test for the fallback showed the fallback was *itself* CE-invalid, on exactly one field: `deviceSupportRisk`. So the CORE/FULL split is back — for precisely the reason it existed before #311 removed it as "no longer needed". Chain is paged → unpaged FULL → unpaged CORE. A degraded page is reported as **unfiltered** whatever was asked for, and the filter controls disable.

**I silently dropped both documents out of the `validate:graphql` gate.** The script skips any `gql` body containing `${...}`; extracting the shared selection set made both dynamic. Measured: `origin/main` had 1 document and 0 skipped; my first commit had 2 and skipped both — including the one the gate had been checking. That *is* the house split (dynamic documents are covered by a spec), but the spec has to exist. `sbomComponentsSchemaDrift.spec.ts` is the replacement, and it asserts the Pro-only documents are **invalid** on CE with the message named — so the day the sync lands, the test fails and tells whoever lands it to delete the fallback.

**Three ordering bugs** needed a request fence, so the paging state moved into `useSbomComponentsPaging`: a Load-more landing after a filter change (appended ALL rows to an UNATTESTED list and corrupted the cursor), two searches resolving out of order, and a trailing debounce after navigation that left release B showing release A's SBOM *permanently*. Each is now a test that fails without the fence.

Plus: release-change state reset, unmount disposal, failed-filter-change clearing, the filtered empty state no longer swallowing the tree view (where nothing on screen could clear the filter), labels keying off the applied filter, a null-cursor `loadMore` guard, and filter changes no longer nuking the dependency-graph cache.

## Validated against the running backend

Not just unit tests — hot-swapped `rearm-core` built from `fda-readiness-1-dev@7303cb29` into the sandbox and probed it with seeded data:

- release-scoped gauge total **7 == page `ALL` total 7**, root excluded from both;
- `ATTESTED` 2 + `UNATTESTED` 5 == `ALL` 7, and `gauge.attested == ATTESTED`;
- cursor walk at limit 3 returned `totalCount` 7 on **every** page, visiting 7 items / 7 distinct;
- **walking `UNATTESTED` while attesting each page** visited all 5 remaining exactly once and closed the gauge to 7/7 — the walk offset pagination could not do;
- `search: "lib%"` → 0 while `"lib0"` → 7, confirming `ESCAPE '!'` literal handling in a real deployment.

## Checks

- `npm run test`: **234 tests, 233 pass**. The one failure is `routeInputSchemaDrift.spec.ts`, pre-existing and identical on clean `origin/main`. Worth stating plainly since **nothing in CI runs vitest** — this count is a manual gate.
- `npm run lint`: 273 problems with *and* without this diff.
- `npm run validate:graphql`: 3 Pro failures, identical on `origin/main`, none in these files.
- `vite build` passes. Added lines ASCII-clean (detector validated against a planted em-dash). Note this repo has no `vue-tsc`, so the build is the type gate rather than a full SFC typecheck.

### DIFF
--- a/ui/src/components/ReleaseView.vue
+++ b/ui/src/components/ReleaseView.vue
@@ -917,10 +917,10 @@
                                 v-if="sbomViewMode === 'list'"
                                 v-model:value="sbomSearchQueryInput"
                                 @update:value="onSbomSearchInput"
-                                @clear="onSbomSearchInput"
                                 placeholder="Search by canonical purl"
                                 clearable
                                 size="small"
+                                :disabled="sbomDegraded"
                                 style="width: 320px;"
                             />
                             <n-select
@@ -929,6 +929,7 @@
                                 @update:value="onSbomFilterChange"
                                 size="small"
                                 style="width: 260px;"
+                                :disabled="sbomDegraded"
                                 :options="sbomAttestationFilterOptions"
                             />
                             <n-radio-group v-model:value="sbomViewMode" size="small" @update:value="handleSbomViewModeChange">
@@ -964,14 +965,20 @@
                             <n-spin size="medium" />
                             <p>Loading SBOM components...</p>
                         </div>
-                        <div v-else-if="sbomComponentsLoaded && sbomComponents.length === 0">
+                        <!-- Scoped to the LIST view. A filtered-empty result used to sit
+                             ahead of the tree branch in this chain and swallow it, and the
+                             search and filter controls are list-only, so the operator was
+                             left staring at "no components match" with nothing on screen
+                             able to clear the filter. -->
+                        <div v-else-if="sbomComponentsLoaded && sbomComponents.length === 0
+                            && sbomViewMode === 'list'">
                             <!-- "the filter matched nothing" and "the release has no
                                  components" are different facts, and the second one is a
                                  claim about the SBOM. Saying it while a filter is active
                                  would report an empty inventory for a release that has one. -->
-                            <p v-if="sbomAttestationFilter !== 'ALL' || sbomSearchQueryInput">
+                            <p v-if="sbomFilterIsActive">
                                 No components match this filter.
-                                <template v-if="sbomAttestationFilter === 'UNATTESTED'">
+                                <template v-if="sbomAppliedFilter === 'UNATTESTED'">
                                     Every component in this release has a support disclosure.
                                 </template>
                             </p>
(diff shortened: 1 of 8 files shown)


logicflakes:
Closing: wrong branch shape on my part, not a content problem.

This was opened `fda-readiness-1-dev -> main`, which is wrong twice over. It would put the FDA UI into CE `main` before the schema sync (PR 6) — the memo's order is backend sync first, then UI, and "merge to CE main" is the delivery step, not the review unit for 5.1. And because the head is a long-lived branch, every later commit to it would silently grow this PR — the same trap as relizaio/rearm-saas#481.

Re-opening as `2026-09-7e-ui-paged-list -> fda-readiness-1-dev` with the same content and both layers' findings. Steps 5.2-5.4 follow the same shape, and the eventual `fda-readiness-1-dev -> main` PR gets opened only after PR 6.

PR #323, opened 2026-09-05
FDA-Readiness-1 PR 5.4/4: bulk support attestation over the current filter

logicflakes:
Last of the four PR-5 UI slices. Adds "Attest all shown": walk the currently applied
filter, confirm the exact set, then write the same attestation to every component in
batches.

Stacked on `fda-readiness-1-dev` (5.1 paged list, 5.2 coverage gauge, 5.3 attestation
form). CE sync is PR 6, separately.

## What it does

- **Collect, then write. Never interleaved.** The walk pages at `limit: 500` and writes
  nothing; batches of 200 go out only after the operator confirms. Each `APPLIED` shrinks
  the `UNATTESTED` set under the cursor, so a walk that also writes cannot be trusted
  about its own remainder.
- **Confirmation stage.** A count alone cannot catch a filter that silently reset, so it
  names the exact count, the filter and search the walk actually ran under, ten components
  by name, and every value about to be written -- including `party`, which is exported on
  the BOM, and the audit reason, which is not.
- **Refuses rather than truncates.** Over 5,000 ids it refuses with the real total. A
  degraded (CE-fallback) page is refused outright: a degraded page is unfiltered, so
  sweeping it would attest the whole release under a filter label that is a lie.
- **A write never degrades.** CE takes 4 mutation arguments to Pro's 12; a fallback would
  drop the level and the justification while reporting success. Schema drift is a refusal.
- `reason` is mandatory, `assessedAt` is one instant captured at confirmation, and a
  negative level (`NO_LONGER_MAINTAINED` / `ABANDONED`) needs a justification -- in a
  sweep that one basis is asserted about every component selected.
- `SKIPPED_ATTESTED` on an `UNATTESTED` walk is surfaced, not hidden: those components
  were undisclosed when the sweep started, so a colleague recorded them while it ran.

## Gate

Both layers ran on the settled diff and the findings are fixed in `7fd40973` (see the
commit body for the full list). The load-bearing ones:

- The form stayed editable through the walk and the Attest button never re-validated, so
  clearing **Reason** mid-walk wrote thousands of attestations with no audit reason --
  `reason` is a UI-only rule the server does not enforce.
- **Esc dismissed the modal mid-sweep.** naive-ui checks `closeOnEsc` independently of
  `closable`/`maskClosable`, so locking those two was not enough. The sweep ran on into a
  closed modal and its only report was lost.
- An aborted sweep rendered as a green "0 attested.", and "re-running is safe" was
  appended to schema drift, where a retry can never succeed. `BulkOutcome.retryable` now
  carries the distinction.
- `unknownOutcomes` was computed as a safety signal and never rendered.
- The confirmation and the write read the filter live; the search debounce and a degraded
  page can both move it behind the open modal.

**Validation.** 417 specs (416 pass; the one failure, `routeInputSchemaDrift`
`notifyComponentOwner`, is pre-existing on `fda-readiness-1-dev` and untouched by this
diff -- worth a separate look). Every new wiring assertion was revert-probed; one passed
against the reverted code and was tightened, and the spec helper silently matched only
`async function`, so every assertion about `openBulkAttest` had been running against an
empty string.

Browser-probed on the sandbox, because `ui/` has no `vue-tsc` and a green build proves
nothing about the DOM: form frozen during the walk (0 disabled controls before, 5
during), Esc blocked with the sweep genuinely in flight, and both done-stage branches
rendered -- amber "0 attested." with no retry advice on drift, the unknown-outcome alert
on a mixed batch. Zero page exceptions. The source scans passed the Esc fix; the browser
showed it firing at the wrong moment.

**Recommend `/code-review ultra` before merge** -- this writes a regulatory claim across
thousands of components on one click, with no bulk undo. I cannot launch it.

## Follow-up, already filed

`t20260905-134808-29907` (rearm-saas, before PR 6): server-minted `batchId` (V84 column +
optional arg + response field), a server-side 1,000-id cap, and `findAllById` -> array-bound
lookup at `SbomComponentService:1367`.

### DIFF
--- a/ui/src/components/ReleaseView.vue
+++ b/ui/src/components/ReleaseView.vue
@@ -501,6 +501,7 @@
                 style="width: 720px;"
                 :mask-closable="!bulk.submitting.value"
                 :closable="!bulk.submitting.value"
+                :close-on-esc="!bulk.submitting.value"
                 title="Attest every component matching this filter">
                 <!-- COMPOSE -->
                 <div v-if="bulkStage === 'compose'">
@@ -510,7 +511,11 @@
                         withdrawing each component individually, and a withdrawn claim stays
                         on the record.
                     </n-alert>
-                    <n-form label-placement="top">
+                    <!-- Frozen during the walk: the collect is awaited over many round
+                         trips while this stage stays mounted, so an edit made after the
+                         click would be written by a confirmation the operator already
+                         approved against different values. -->
+                    <n-form label-placement="top" :disabled="bulk.collecting.value">
                         <n-form-item label="Level of support">
                             <n-select v-model:value="bulkForm.levelOfSupport"
                                 :options="LEVEL_OPTIONS" placeholder="Not stated" />
@@ -555,9 +560,9 @@
                             components will be attested.
                         </div>
                         <div style="font-size: 12px; margin-top: 6px;">
-                            Filter <strong>{{ sbomAppliedFilter }}</strong><span
-                                v-if="sbomAppliedSearch">, search
-                                <strong>"{{ sbomAppliedSearch }}"</strong></span>.
+                            Filter <strong>{{ bulkAppliedLabel }}</strong><span
+                                v-if="bulkCollected.search">, search
+                                <strong>"{{ bulkCollected.search }}"</strong></span>.
                             There is no bulk undo.
                         </div>
                     </n-alert>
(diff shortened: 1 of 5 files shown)


logicflakes:
Follow-up applied in `c4c424f8`.

`submit` now calls `validateBulkInput` and refuses on a violation, returning an outcome carrying the message with `aborted: true, retryable: false`. That covers the negative-level basis rule as well as the reason, so the composable enforces its own invariant rather than trusting the caller's disabled-button binding.

Two new specs, both revert-probed (they fail with the guard removed). **Four existing specs were calling `submit` with no reason and passing** -- which is the gap itself, stated in test form; they now supply one.

419 specs, 418 pass. The remaining failure is the pre-existing `routeInputSchemaDrift` / `notifyComponentOwner` one, present on `fda-readiness-1-dev` and untouched by this diff (filed separately).

PR #325, opened 2026-09-06
helm: require the oci-registry secret instead of silently starting without it

logicflakes:
Found while unblocking the FDA acceptance test, which needed a working BOM ingest. Not
FDA-related, so it targets `main` directly rather than the feature branch.

## What happens today

The oci-artifact service exits fatally when `REGISTRY_USERNAME`, `REGISTRY_TOKEN` or
`REGISTRY_HOST` is empty (`REGISTRY_USERNAME or REGISTRY_TOKEN or REGISTRY_HOST not set`,
`main.go:322`). **It has no anonymous mode**, so there is no configuration in which an absent
`oci-registry-secrets` is survivable.

The chart marked both `secretKeyRef`s `optional: true`. A missing Secret therefore does not
stop the pod starting -- it starts with empty credentials, dies, and keeps dying.

I found this in a live deployment at **1560 restarts**, showing only `CrashLoopBackOff`. The
real cause (`create_secret_in_chart: false` with no externally-managed Secret) was visible
nowhere except the container log.

The consequence was not limited to that service. **BOM ingest was down the whole time**:
rebom validates and dedups the BOM fine, then cannot push it, and `addBom` fails. What
actually reached the user on upload was:

```
BOM processing failed: Rebom error: fetch failed
```

-- four services away from the fault, naming none of it.

## The change

Drop `optional: true`. Kubernetes then reports the fault on the pod, in the words of the thing
that is actually wrong:

```
CreateContainerConfigError: secret "oci-registry-secrets" not found
```

`values.yaml` now states the contract at `create_secret_in_chart`, where someone choosing that
value reads it: `false` means you supply the Secret yourself, there is no anonymous mode, and
`bundledRegistry` is the option if you have no external registry.

## Validation

Both paths rendered with `helm template`:
- external path emits required refs to `oci-registry-secrets`
- `bundledRegistry.enabled=true` still resolves to `rearm-zot:5000` with `rearm-zot-credentials`

`helm lint` clean.

## Note for anyone hitting this

If you are already in the crashloop, the quickest route out is often the bundled registry --
`ociArtifactService.bundledRegistry.enabled: true` -- which needs no external registry or
credentials. Adding it to an existing deployment requires one deploy with `--force-conflicts`,
as `values.yaml` documents.


logicflakes:
Closing. The helm fix is not FDA-related -- this PR says so itself -- but the
branch was cut from the FDA work, so it carried 15 commits / 89 files / 13.5k
insertions of FDA-Readiness attestation UI and a CE backend sync along with the
19-line chart change, and conflicted on `ReleaseView.vue`.

**The helm fix lives on in #366**, re-cut onto `main` with the original commit
and authorship preserved, and all three render paths re-verified there (lint,
external, default, bundled). Nothing is lost by closing this.

**The FDA commits are not lost either** -- `96d2e948`, `62025485`, `641ef257`,
`10c30e28`, `ab8daa68` and the drift-fallback refactors are already on the
`origin/2026-09-7*` branches, which is where that work is actually landing.

**One thing here does still need a home:** `040311f8` ("test(ui): CE now accepts
notifyComponentOwner -- retarget the assertion"). `main`'s UI suite is red, and
that commit fixes one of the four failures -- the assertion that CE lacks
`notifyComponentOwner`, which stopped being true once the field reached the CE
schema (`schema.graphqls:1742`). It cherry-picks onto `main` on its own.

For the record, since I had this wrong earlier: that commit is **not** the whole
of main's redness. The other three failures are
`Cannot query field "recognized"/"attestation" on type "SourceCodeEntry"` -- the
UI querying fields the CE schema does not have, introduced by `92b820f1`
(integrity / commit claims, #363) after this branch was cut. Unrelated to FDA and
unrelated to this PR; it needs a CE schema sync of its own.

PR #326, opened 2026-09-06
feat(instance): filterable target version picker; edit extra configuration after deployment

taleodor-claude:
Two changes on the instance page:

1. **Set Target Release**: the release dropdown is filterable, so the version can be typed to narrow the list.
2. **Extra configuration after deployment**: the Config column of a linked feature set showed a bare contenteditable span (easy to miss, easy to change by accident). It now shows the value with an edit icon that opens a modal (textarea, Submit, Reset). Saving writes a plan revision like every other plan change, so it is recorded in the Plan History tab, which gains a **Configuration** change type filter and label.

The Configuration change type is classified by the backend (Pro backend change shipping separately). Until that backend is deployed, picking the Configuration filter is unknown to the server; everything else works against the current backend.

Verification: `vite build` clean; eslint errors in `InstanceView.vue` are pre-existing on main (56 on main, 55 on the branch, all in an untouched block).

### DIFF
--- a/ui/src/components/CreateRelease.vue
+++ b/ui/src/components/CreateRelease.vue
@@ -45,6 +45,8 @@
                         v-if="releaseCandidates && releaseCandidates.length"
                         v-model="release.uuid"
                         v-on:update:value="value => checkIfCreateNewRelease(value)"
+                        filterable
+                        placeholder="Select or type to filter by version"
                         :options="releaseCandidates" />
                 <div v-if="!isReleasesLoading && releaseCandidates && releaseCandidates.length === 0" style="color: #999; font-style: italic;">
                     No releases discovered.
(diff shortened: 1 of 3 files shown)


taleodor-claude:
Verified live on psclaude with this branch's UI build and the matching backend change, on a standalone instance with one TARGET product (`probe-instance-config-edit.js`):

- Config cell: edit icon opens the modal; saved value shows immediately and survives a reload (plan revision 41), restore to the original value writes revision 42.
- Plan History: both revisions classify as **Configuration** (backend), and the Configuration filter returns exactly those two.
- Set Target Release: the release select is filterable; typing a full version (`26.02.4.12`) narrows 10 options to 1.
- No GraphQL errors during the run.

Note for reviewers: the Plan History tab defaults to the last 30 days, and a revision's change type is computed against the previous revision in the fetched window, so on an instance whose previous plan revision is older than 30 days the newest row shows no type until the range is widened. Pre-existing behaviour, not changed here.

PR #328, opened 2026-09-07
FDA-Readiness-1 7f step 2: CE backend sync (prose storage, prose bound, SUPPORT_WINDOW enum)

logicflakes:
Mechanical mirror, reproducible rather than trusted:

```
git -C rearm-saas checkout 1296cfd3
bash rearm-saas/backend/copy-src.sh <rearm>/backend
```

Brings rearm-saas #498 (four manufacturer prose slots + the `SUPPORT_WINDOW` enum fix) and
#499 (prose bound lowered to 8,000). Five files, no hand edits.

## Why this lands before the UI

The four `fda*` fields are selected by `store.ts`'s `organizations` query, which runs on
**every page load**, and `fetchMyOrganizations` catches a failure and returns **no org** — so
a CE install running the UI against a CE backend without these fields is *unusable*, not
merely degraded. That is also why no drift fallback was added there: the fix is this
ordering, not a runtime workaround.

## It also closes a page-breaking bug already in CE

Before this, CE's schema documented `SUPPORT_WINDOW` in prose while the enum did not declare
it. Any CE release that ever had its device support window set could not be serialized, and
the **whole** release query failed — the page would not load at all. PR #324 synced from a SHA
predating the fix, so CE carried it.

## Confirmations in the CE tree after syncing

| | |
|---|---|
| `fdaAssessmentNarrative`, `fdaPatchesMayCeaseStatement`, `fdaRiskTransferProcessRef`, `fdaRiskIncreasesNotice` | present on **both** the `Settings` type and `SettingsInput` |
| `SUPPORT_WINDOW` | inside `enum ReleaseUpdateScopeEnum` itself, not merely in a description |
| `eos`, `eol`, `clearEos`, `clearEol` | on `ReleaseInput` |
| `FDA_PROSE_MAX_LENGTH` | `8000` |

## Gates

- **Taxonomy doc**: `git diff --exit-code` clean, and all three copies (source, CE root, CE
  backend) byte-identical.
- **`CyclonedxTaxonomyDocSyncTest`**: 2/2 green *inside the CE build*.
- **`ReleaseUpdateScopeSchemaEnumSyncTest`**: 1/1 green inside the CE build — this is the
  guard that would have caught the bug above.
- **Full CE suite: 1777 tests, 0 failures, 0 errors.** The known
  `anUnreadablePayloadIsExcludedWithoutFailingTheBatch` flake did not recur.

Per operator ruling there is no `dev -> main` for this track until the feature is complete and
demoed, so CE main stays exposed to the forged-provenance hole and to the `SUPPORT_WINDOW`
serialization bug until then.

### DIFF
--- a/backend/src/main/java/io/reliza/model/OrganizationData.java
+++ b/backend/src/main/java/io/reliza/model/OrganizationData.java
@@ -105,6 +105,58 @@ public class OrganizationData extends RelizaDataParent implements RelizaObject {
 		@JsonProperty
 		private Integer notificationRetentionDays;
 
+		/**
+		 * FDA-Readiness-1 section 7f. Manufacturer-authored prose that the generated
+		 * submission documents render. Org-level because all four say something about how
+		 * THIS manufacturer assesses and supports, not about one device.
+		 *
+		 * <p>Nullable throughout, and null is a real state rather than a default to paper
+		 * over: a document with an empty required slot is BLOCKED rather than generated with
+		 * the section missing. On a submission a reviewer expects gaps, but on the
+		 * patient-facing Device Support Statement a silent gap is itself the misleading
+		 * thing under 502(a)(1).
+		 *
+		 * <p>There is deliberately NO product-shipped default text for any of them. These
+		 * are the manufacturer's own commitments; shipping wording they did not write and
+		 * then putting it over their name is the failure this feature exists to avoid.
+		 */
+		@JsonProperty
+		private String fdaAssessmentNarrative;
+
+		/** Statement that patches/updates may cease at end of support (labeling VI.A.2). */
+		@JsonProperty
+		private String fdaPatchesMayCeaseStatement;
+
+		/**
+		 * REFERENCE to the controlled risk-transfer process (labeling VI.A.3) -- an
+		 * identifier or URL, not the process itself. Plans and processes are
+		 * design-history-file records; pasting one into a generated artifact creates a
+		 * second, unversioned copy that drifts from the controlled original.
+		 */
+		@JsonProperty
+		private String fdaRiskTransferProcessRef;
+
+		/** Notice that end-user cybersecurity risk increases over time (labeling VI.A.4). */
+		@JsonProperty
+		private String fdaRiskIncreasesNotice;
+
+		/**
+		 * Bound on each prose field.
+		 *
+		 * <p>8,000 rather than the 20,000 an earlier revision used, and the reason is the
+		 * READ side rather than the write. These live on the organization record, which
+		 * {@code getOrganizationData} materialises fresh on every call -- including from
+		 * {@code AuthorizationService}, i.e. on the authorization path of ordinary requests.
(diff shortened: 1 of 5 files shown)


logicflakes:
Both layers ran. **No blocking findings from either.**

## Independently verified by hashing, not eyeballing

Both layers hashed `backend/src` in both trees rather than reading the diff:

- **881 of 882 CE files byte-identical** to the rearm-saas source at `1296cfd3`
- the one content difference is `VersionAssignmentTest.java`, a deliberate exclusion (CE keeps its own)
- five files exist only in rearm-saas, all six `copy-src.sh` exclusions resolve to real files, **no stale exclusion**
- `git diff --name-status` shows **zero `D` entries** — rsync `--delete` clobbered nothing; CE's unsynced `oss/` tree is intact
- upstream `git diff f491aece..1296cfd3` is exactly the same five files as this PR, so nothing was left behind and nothing rode along

## CE runtime wiring — clean

Settings is not hand-wired on either side: the write path converts a `Map` through `Utils.OM.convertValue(…, Settings.class)` and the read path is DGS's default property fetcher over Lombok getters, so new `@JsonProperty` fields flow through with no per-field code. Neither repo has a `settings` resolver in its non-synced tree. No new cross-tree beans; the four interfaces without CE implementations are all pre-existing and injected `required = false` with explicit null fallbacks.

Layer 1 also checked the CE-specific hazard I would have missed: CE's `TestInitializer` diverges from the SAAS one (it builds the org via the repository rather than `SaasOrganizationService`), so a fresh CE org has **no** settings object at all. `updateSettings` creates one when null, and every test calls it before reading, so nothing NPEs on the CE path.

## The step-3 gate is already satisfied

`validate-graphql`: **316 documents, 0 CE drift warnings.** The four `fda*` warnings are gone. The 3 remaining Pro failures (`defaultView`, `deployedTo`) are pre-existing dead paths — verified absent from the *base* CE schema too, so this sync did not remove them. Reverse drift is structurally impossible here: the diff has **0** removed lines. 395 UI tests pass.

## Layer 1's one caution, answered

It noted the host `target/` was stale, so there was no local artifact proving the two new tests compile in the CE tree — it verified symbol resolution by hand instead. Ran them for real on the host against the shared DB:

```
FdaProseSettingsTest                    9/9
CyclonedxTaxonomyDocSyncTest            2/2
ReleaseUpdateScopeSchemaEnumSyncTest    1/1
```

Plus the full CE suite inside the container build: **1777 tests, 0 failures, 0 errors.**

## One finding, routed upstream deliberately

`FdaProseSettingsTest.java:240-241` uses inline FQNs (`java.util.function.BiConsumer`, `java.util.List.of`), against the imports rule. It is **not** fixed here: CE's copy is generated, so patching it would be erased by the next sync *and* would break the byte-identity property this PR's reviewability rests on. Carried into the step-4 addendum PR (rearm-saas) and recorded on `t20260904-041253-3993`.

Ready to merge.

PR #330, opened 2026-09-07
Author the FDA prose slots and the device support window in the UI (7f)

logicflakes:
Step 3 of 7f: the UI that writes what rearm-saas#498 and rearm#329 made storable. Four
org-level prose slots in Settings, and the device support window on a PRODUCT release.

## Clear semantics, end to end

Prose diffs against a BASELINE captured at load, so an emptied field is sent as `""` and
CLEARS, while a field the operator never touched is OMITTED and left alone. That is the
`blankToNull` contract the backend already implements; sending the whole form every time
would make an unrelated save silently rewrite all four.

## The device window

PRODUCT releases only, and deliberately editable after assembly -- a support window is
declared years after a device ships and revised as it ages, so a DRAFT-only gate would
mean no shipped device could ever have one. Coherence (`eos <= eol`) is NOT pre-validated
here: the rule is enforced for every writer, so echoing the server verbatim avoids
inventing a second wording that could drift.

## Layer 1 found a HIGH bug in this branch; it is fixed

`eos`/`eol` were selected by the product-simplified query only. `fetchRelease` switches to
the FULL query once the Underlying Artifacts tab has been visited, so after that the
refetch following a successful save read `undefined`, blanked both pickers AND reset the
baseline -- leaving the window looking permanently unsaveable while the value sat safely
in the database. Retyping and saving again did nothing visible, because the backend
correctly saw no change and emitted no event.

Fixed by selecting the fields in BOTH queries, and by making `seedDeviceWindow`
distinguish `undefined` (this query did not ask) from `null` (the server says not
declared). The second guard should be unreachable given the first; it is there because it
was not unreachable an hour ago.

## Live probe on the sandbox, against the merged backend

Rebuilt CE from `32505f3a` (post-#329) and hot-swapped both deployments.

    W0-W6   card renders on a PRODUCT release, writable, seeded from the server
    D1-D2   both pickers accept a date, Save enables only when dirty
    P1      both dates persist across a reload
    P2      clear EOS alone -- EOL survives
    P3/P3b  server coherence error shown verbatim, and NOTHING was written
    P4/P5   a SUPPORT_WINDOW event row renders the window, not a blank cell
    P6      clear EOL alone -- EOS survives  (the mirror of P2, never exercised before)
    P7      eos and eol both appear in the release network traffic
    P8a-e   THE LAYER 1 PATH: visit Underlying Artifacts, return, set a date, save,
            reload -- value persists and the pickers stay populated
    X       0 page exceptions

P8 was REVERT-PROBED: rebuilt the UI from `a4f5e259` (the commit before the fix) and
re-ran. **P8d fails there (`""` instead of `"2032-12-31"`) while P1-P7 all still pass** --
the entire pre-existing suite was blind to this. P8e passes even unfixed, which confirms
the diagnosis precisely: the data always reached the database, only the display after the
tab round-trip was broken.

Prose probe: T0 seeded, A cleared one field, A2 left the other three untouched, B0/B an
unrelated setting save left all four intact. 0 GraphQL errors, 0 page exceptions.
Strip matrix re-run for regression: 19/19 across all four egresses.

## Checks

`validate-graphql`: **0 CE drift warnings** (3 Pro failures are pre-existing dead paths,
tracked separately). Build clean. **395 tests / 31 files** green -- note nothing in CI runs
the vitest suite, so that count is part of this report by necessity.

## oss/ port decision

This PR is UI-only and touches no `service/oss/` code, so the fork-port question the
sync checklist raises does not arise here. It was answered for this track in #329.

### DIFF
--- a/ui/src/components/OrgSettings.vue
+++ b/ui/src/components/OrgSettings.vue
@@ -944,11 +944,9 @@
                              are the manufacturer's own commitments, and words nobody here
                              wrote going out over their name is the failure the whole
                              feature exists to avoid. -->
-                        <n-divider style="margin: 22px 0 10px;">
-                            <span style="font-size: 13px;">FDA submission and labeling text</span>
-                        </n-divider>
-                        <n-alert type="default" :show-icon="false"
-                            style="font-size: 12px; margin-bottom: 14px; max-width: 760px;">
+                        <n-divider />
+                        <h6>FDA submission and labeling text</h6>
+                        <p class="text-muted" style="max-width: 760px;">
                             Written once here, rendered into every generated document. A
                             document whose required text is missing is <strong>not
                             generated</strong> rather than generated with the section
@@ -957,11 +955,12 @@
                             To remove text, clear the box and save &mdash; an emptied field
                             is recorded as deliberately blank, and a field you do not touch
                             is left exactly as it was.
-                        </n-alert>
+                        </p>
 
                         <n-form-item label="Assessment justification (submission)">
                             <div style="display: flex; flex-direction: column; width: 100%;">
                                 <n-input v-model:value="orgSettings.fdaAssessmentNarrative"
+                                    :disabled="savingOrgSettings" :maxlength="FDA_PROSE_MAX_LENGTH" show-count
                                     type="textarea" :rows="5" style="max-width: 760px;"
                                     placeholder="How components were assessed, and why an upstream end-of-support date is unavailable for most of them." />
                                 <span class="text-muted" style="margin-top: 4px; max-width: 760px;">
@@ -976,6 +975,7 @@
 
                         <n-form-item label="Patches may cease at end of support (labeling)">
                             <n-input v-model:value="orgSettings.fdaPatchesMayCeaseStatement"
+                                    :disabled="savingOrgSettings" :maxlength="FDA_PROSE_MAX_LENGTH" show-count
                                 type="textarea" :rows="3" style="max-width: 760px;"
                                 placeholder="After end of support, security patches and software updates may no longer be provided." />
                         </n-form-item>
@@ -983,6 +983,7 @@
                         <n-form-item label="Risk-transfer process reference (labeling)">
                             <div style="display: flex; flex-direction: column; width: 100%;">
                                 <n-input v-model:value="orgSettings.fdaRiskTransferProcessRef"
+                                    :disabled="savingOrgSettings" :maxlength="FDA_PROSE_MAX_LENGTH" show-count
                                     style="max-width: 760px;"
                                     placeholder="e.g. DHF-PROC-4471 rev C, or a URL to the controlled document" />
                                 <span class="text-muted" style="margin-top: 4px; max-width: 760px;">
@@ -995,6 +996,7 @@
 
                         <n-form-item label="Risk increases over time (labeling)">
                             <n-input v-model:value="orgSettings.fdaRiskIncreasesNotice"
+                                    :disabled="savingOrgSettings" :maxlength="FDA_PROSE_MAX_LENGTH" show-count
                                 type="textarea" :rows="3" style="max-width: 760px;"
                                 placeholder="Cybersecurity risk to users can be expected to increase after end of support." />
                         </n-form-item>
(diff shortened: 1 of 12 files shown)


logicflakes:
## Both layers reported; all findings fixed at `adf8a063`

### Layer 1 -- HIGH, confirmed by tracing it end to end

`proseBaseline` was refreshed only by `loadOrgSettings()` nested INSIDE the save's `try`,
and that helper does an unguarded network round trip. When it rejects after the mutation
has committed: the operator is told **"Save Failed" about a save that succeeded**, and the
baseline stays stale. They clear the box to start over -> `proseDiff` compares `''` against
the stale `''` -> **field omitted, not cleared** -> "Settings Saved". The operator believes
the narrative is blank; the server still holds the text; the generator renders it into a
patient-facing document. That is the fabricated-prose failure this feature exists to
prevent, reached without a single unusual step.

Both forms now take the baseline from the **mutation's own response**, before anything that
can throw. It was already in hand -- the window save was selecting `{ uuid eos eol }` and
throwing it away. The refetch moved outside the `try`, so its failure reports
"Settings were saved, but reloading them failed" instead of masquerading as a failed write.

Same shape on the window with a worse ending: `fetchRelease` can throw, and with a stale
baseline on a committed write the user could not even clear the date they had just set --
dirty `false`, Save disabled, nothing explaining why.

Also fixed: prose inputs stayed editable during a save (post-click keystrokes overwritten
under a green toast); no length cap, so a 9000-char paste failed the **whole** settings form
atomically and lost unrelated edits; `parseGraphQLError(err.message)` threw a `TypeError`
out of the `catch` on a rejection without a string `.message`; a failed save left a red
alert with no path to clearing; and the comment claiming "no server-side lifecycle gate"
was **wrong** -- the resolver calls the `DRAFT_ONLY` overload, it just never fires because
`isAssemblyRequested` trips only on fields this payload omits. Right conclusion, wrong
reason, and the real reason is load-bearing: widening the payload ARMS that gate.

### Layer 2 -- verdict "written by someone who knows this codebase"

Its main finding was fair and I had earned it: `proseDiff`'s own comment named its
precedent, `utils/supportAttestationInput.ts`, and that precedent has two spec files. I
copied the shape but not the part that mattered. Three modules extracted
(`fdaProseInput`, `deviceSupportWindowInput`, `supportWindowDisplay`), 35 assertions.
Also dropped a no-op `as any`, converted the only nine GraphQL `#` comment lines in
`graphqlQueries.ts` to this file's `//`-above-the-const style, and switched the new
sub-section to the `n-divider` + `h6` + `p.text-muted` pattern the sid-PURL section uses.

### Then the extraction broke the page, and only the browser caught it

Deleting the `FDA_PROSE_FIELDS`/`proseDiff` block took the `proseBaseline` declaration with
it. **Build passed, 427 unit tests passed, eslint reported no new errors** -- and every
prose save threw `proseBaseline is not defined`. The live probe went green -> T0/A2/B
failing with 5 page exceptions. No vue-tsc, and `no-undef` does not apply to
`<script setup>`, so nothing in the toolchain had anything to say. Note what this means:
extracted logic having its own passing specs says nothing about whether the component still
references it.

`fdaProseWiring.spec.ts` closes that, in the style of `releaseViewBulkWiring.spec.ts`. It
asserts the four inputs are capped and frozen **as a set**, so adding a fifth field and
forgetting one fails, and it asserts both save paths refresh the baseline before the
refetch -- the HIGH, guarded by ordering rather than by comment.

### Revert-probes (three)

- delete `const proseBaseline` -> `declares proseBaseline` FAILS
- remove `eos`/`eol` from the full query -> `SingleReleaseGql selects the device support
  window` FAILS while the product query still passes. **That asymmetry is the shipped bug**,
  and a comment was previously the only thing holding the pair in step -- both documents
  interpolate `${...}`, so `validate-graphql` skips them and does not even count the skip.
- add a stray `artifacts` key -> `never a dead field` FAILS. `updateRelease` treats a null
  list as no change, so that assertion guards against a widening that would DETACH contents.

### Final state

**448 tests / 35 files** (was 395/31). `validate-graphql` **0 CE drift**. eslint no new
errors. **Zero non-ASCII bytes added.** Live re-run on the rebuilt image: prose
T0/A/A2/B0/B green, window W0-W6 / D1-D2 / P1-P8e green, **0 page exceptions on both**.

PR #331, opened 2026-09-07
The FDA support addendum, rendered from one client-side collector (7g)

logicflakes:
Step 4b. Pairs with rearm-saas#500, which adds the per-release narrative and the resolution
seam this reads through.

## Why the browser renders it

The PDF and the Device Support Statement are frontend-rendered by the plan, so **one
client-side assembly feeds all three documents**. A backend-rendered CSV would be a second
source of truth for the same document -- two codebases computing "how many components have
no attestation".

It also could not have reused the existing CSV path even if we wanted to: that proxies to
the separate **rebom** service (`ReleaseService.java:879`), which only ever sees the merged
BOM -- and the strip work guarantees that document carries no support properties at all
(matrix EGRESS 4 asserts "CSV carries no reliza properties"). The addendum is made entirely
of facts rebom cannot see.

## The collector is deliberately document-agnostic

`addendumData.ts` returns FACTS, never rendered text. `deviceEos` is a date string or null,
never `"not declared"` -- that phrase is something the CSV *says*, and a PDF may word it
differently. A spec asserts no renderer vocabulary leaks into the collected data, so the
next PR inherits raw values rather than a spreadsheet's phrasing. **The next PR consumes
this unchanged.**

## It refuses rather than truncating

A partial addendum is the dangerous output: a regulatory document that looks complete while
under-reporting how many components have no support attestation -- the one number a reviewer
is looking for. Every refusal returns **no data at all** rather than data plus a flag, the
same rule the bulk sweep follows, so a caller cannot destructure past the flag.

It refuses above the sweep's ceiling, on an empty page that claims more, on a query failure,
and when the collected row count disagrees with the gauge. That last one matters: they are
two reads of the same scope moments apart, so a mismatch means the BOM changed underneath
the walk and the document would state a count its own rows contradict.

**No narrower fallback, deliberately.** A degraded document would be missing exactly the
columns the addendum exists to carry, and would read as an unassessed release. Same walk
limit and ceiling as bulk attest -- same scope, same server, so a release the sweep refuses
this must refuse too. Counts come from the gauge's own loader, not a second copy of its
query, so the document and the page cannot disagree.

## The CSV writer

Hand-rolled rather than a dependency: the surface is one escaper, and the cells carry up to
8,000 characters of operator prose. RFC 4180 quote-doubling, embedded newlines, CRLF, and
`=`/`+`/`-`/`@` neutralised -- formula injection is not theoretical when the cell is free
text an operator typed, and a justification that renders as a hyperlink means the exported
evidence does not say what the record says.

**Three revert-probes**: dropping quote-doubling fails 3, dropping neutralisation fails 6,
narrowing the quote trigger fails 4. The round-trip test parses with a deliberately naive
RFC-4180 reader, because a writer that only satisfies a lenient parser is wrong.

## Document decisions worth reviewing

- **Level and justification are alternatives**, per L1021-1022 -- the justification exists
  *because* there is no level to give, so emitting both would suggest it qualifies the level.
- **WITHDRAWN reads as not assessed** and its stale dates are suppressed: a withdrawn
  attestation is not injected into exports, so counting it would make the addendum disagree
  with the BOM it accompanies.
- **Absent facts say "not assessed"**, not a blank cell -- blank reads as something the
  exporter forgot rather than as "we looked and have nothing recorded".
- **Device EOS and EOL are two separate labelled facts.** One merged "support window" line
  forces the reader to guess which a lone date is.

## The BOM would have hard-failed CI

`csvDocument` prepends a UTF-8 BOM so Excel decodes the file on a double-click open --
without it any non-ASCII component name opens as mojibake. It was a **literal U+FEFF in the
source** until the ASCII sweep caught it: exactly the invisible character the CI check exits
123 on, before maven runs. Escaped to `'﻿'`, with the reason in a comment so it does
not come back.

## Checks

- **516 tests / 39 files** (was 448/35). Build clean.
- **One CE drift warning, expected and unfixed by design**: the release query selects
  `fdaAssessmentNarrative`, which CE gains at the deferred sync. Per the ruling drift is
  informational and there is deliberately no fallback.
- Zero non-ASCII bytes added.
- `addendumExportWiring.spec.ts` continues the `*Wiring.spec.ts` pattern -- no vue-tsc here,
  and on the previous PR that cost a working feature while build, tests and eslint were all
  green.

### DIFF
--- a/ui/src/components/ReleaseView.vue
+++ b/ui/src/components/ReleaseView.vue
@@ -112,8 +112,22 @@
                             <n-radio-button value="JSON">CycloneDX 1.6 (JSON)</n-radio-button>
                             <n-radio-button value="CSV">CSV</n-radio-button>
                             <n-radio-button value="EXCEL">EXCEL</n-radio-button>
+                            <!-- Its own export TYPE, not a toggle on the BOM export. The
+                                 addendum is a different document that happens to share this
+                                 modal: it is assembled in the browser from the support
+                                 attestations and the org prose, and the BOM-shaping options
+                                 below do not apply to it. A checkbox would imply it rides
+                                 along with whichever format was picked. -->
+                            <n-radio-button value="FDA_ADDENDUM">FDA support addendum (CSV)</n-radio-button>
                         </n-radio-group>
                     </n-form-item>
+                    <n-alert v-if="selectedSbomMediaType === 'FDA_ADDENDUM'" type="default"
+                        :show-icon="false" style="font-size: 12px; max-width: 620px; margin-bottom: 10px;">
+                        Every component in this release with its level of support, the end-of-support
+                        date where one is attested and the justification where none is, plus the
+                        device support window and the assessment justification. The counts come from
+                        the coverage gauge, so this document and the page always agree.
+                    </n-alert>
                     <n-form-item v-if="selectedSbomMediaType === 'JSON'">
                         <template #label>
                             <span style="display: inline-flex; align-items: center;">
@@ -1685,6 +1699,8 @@ import gql from 'graphql-tag'
 import graphqlClient from '../utils/graphql'
 import { GET_VEX_PROPOSALS_BY_RELEASE } from '@/graphql/vexImport'
 import commonFunctions, { SwalData } from '@/utils/commonFunctions'
+import { collectAddendumData } from '@/utils/addendumData'
+import { renderAddendumCsv, addendumFileName } from '@/utils/addendumCsv'
 import { formatSupportWindow } from '@/utils/supportWindowDisplay'
 import { deviceWindowVariables } from '@/utils/deviceSupportWindowInput'
 import graphqlQueries from '@/utils/graphqlQueries'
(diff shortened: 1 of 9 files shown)


logicflakes:
## Both layers reported, all findings fixed, and the live probe found one neither could

Head `de1e7b28`.

### Layer 1 HIGH -- the walk was unbounded, un-deduped, and trusted the cursor

The ceiling only ever compared the server's self-reported `totalCount`, latched from the
first page. Rows actually collected were never bounded. Layer 1 **proved the consequence with
a throwaway spec**: a client returning the same `endCursor` with one item and `hasMore: true`
was walked 200 times, and because the duplicated row count happened to equal `coverage.total`,
the reconciliation **passed** -- `ok: true`, a regulatory document listing one component 200
times. With a non-matching total the same input never terminates: frozen tab, spinner stuck
forever because the `finally` is never reached.

Three guards, each catching a different mask of it: dedupe by `sbomComponentUuid`, a bound on
rows actually **held**, and a refusal when the cursor does not advance. Revert-probed --
removing all three fails exactly the two new tests while every pre-existing refusal test still
passes, which is the point: none of them covered this.

### Layer 1 MEDIUM -- the table contradicted its own header

A level is **optional** on an attestation, and the backend coverage numerator counts rows
where `levelOfSupport IS NULL`. A justification-only attestation is precisely the L1021-1022
case this document exists to carry -- and it rendered as "not assessed" while the header
counted it as attested. Ten such components produced "10 with a support attestation" above
ten rows each reading not assessed. `LEVEL_NOT_STATED` now says what is true.

Also: a WITHDRAWN component emitted its justification prose, **republishing a claim the
manufacturer had formally withdrawn** in the one document where that matters most.

### Layer 2 -- my ceiling rationale was factually wrong

"Same scope, so a release the sweep refuses this must refuse too" -- but the sweep walks the
operator's filter and search while this always walks `attestation: ALL`. On a
6,000-component release with 4,000 unattested the sweep proceeds and this refuses, the
opposite of the claim. Now its own constant with its own reasoning.

Layer 2 also caught that `ADDENDUM_PAGE_QUERY` had **no static validation at all** -- it
interpolates, and `validate-graphql` skips such bodies *before* incrementing its skipped
counter, so it was invisible to every gate. `addendumSchemaDrift.spec.ts` closes it, and
asserts the CE gap is still present so that expected, temporary state fails loudly when the
deferred sync lands.

### Then the live probe found what neither layer could

`ADDENDUM_ORG_QUERY` used `organization(orgUuid:)`. That field is **declared in the schema
but has no resolver** -- it always returns null, and appears under "Unmapped fields" in the
DGS schema report at startup. Every export refused with "The organization could not be
loaded".

Nothing static could have caught it: the field exists, so `validate-graphql` passed, the new
drift spec passed, and the unit specs mocked a client that returned whatever the fixture
said. **A document can be perfectly valid against the schema and still be answered by
nobody.** The codebase already knew -- `graphqlQueries.ts:526` says it in one line, three
files from the one I was writing.

The refusal behaved exactly as designed: it stopped a broken document rather than emitting
one with the labeling statements silently missing, and named the reason precisely enough to
find the bug in one step.

### Pro-backend probe, green end to end

Built `rearm-pro:t3993-7g` from rearm-saas#500, deployed, exported release `1.4.5-demo`:

```
gauge on the page          1 of 10 components have a support disclosure
A9  ROW COUNT == GAUGE     10 == 10
A10 header states the same total (10)
A11 unassessed             9
A12 device EOS and EOL as two separate rows
A13 justification block present, scope stated as "organization default"
A14 every data row carries exactly 8 fields
A5/A6 UTF-8 BOM, CRLF
A2/A3 offered as its own export type; BOM-shaping controls hidden for it
X   0 page exceptions
```

The generated file, verbatim from the download:

```
FDA software support addendum,,,,,,,
Organization,Claude2,,,,,,
Device,walkthrough-demo-service,,,,,,
Release,1.4.5-demo,,,,,,
...
Components in this release,10,,,,,,
Components with a support attestation,1,,,,,,
Components with no support attestation,9,,,,,,
Assessment justification,narrative 1788784329455,,,,,,
Justification scope,organization default,,,,,,
Component,Version,PURL,Level of support,End of support,Justification,Attestation state,Last assessed
debian:openssl,1.1.1n,pkg:deb/debian/openssl@1.1.1n,no longer maintained,2027-06-30,,ATTESTED,2026-09-07T04:28:17Z
```

**541 tests / 40 files** (was 516/39). 0 non-ASCII added. Drift unchanged at 1, expected.

PR #333, opened 2026-09-07
feat(component): Latest tab as the default view; retire the Pull Requests tab

taleodor-claude:
**Latest tab (new default)** on component and product pages: one row per branch / feature set with its newest release: version link, lifecycle, created date, and the same scan status and vulnerability / violation circles as the home page's Most Recent Releases widget. Newest first; clicking a row selects the branch so the right-hand pane follows, as in the Branches tab. An "archived included" switch adds archived branches. Products, which had a plain feature-set table, get the same two tabs.

**Pull Requests tab removed.** It listed branches whose *name* starts with `pull/` or `pullrequest/`, a legacy pipeline convention no current integration produces (Azure DevOps and GitHub builds run on the PR source branch and create first-class Pull Request entities). Such branches now appear in the Branches tab with a **PR** badge, so nothing becomes unreachable.

Backend: needs the `latestReleasesOfComponentPerBranch` query (shipping in the Pro backend; CE gets it with the next sync). Against an older backend the tab shows a message instead of breaking the page.

`vite build` clean; eslint on `ComponentView.vue` unchanged from main (two pre-existing errors).

### DIFF
--- a/ui/src/components/BranchView.vue
+++ b/ui/src/components/BranchView.vue
@@ -377,6 +377,7 @@ import constants from '@/utils/constants'
 import { ReleaseVulnerabilityService } from '@/utils/releaseVulnerabilityService'
 import VulnerabilityModal from '@/components/VulnerabilityModal.vue'
 import { isDtrackConfiguredForOrg, getReleaseScanStatus } from '@/utils/releaseScanStatus'
+import { renderVulnerabilityCells, renderViolationCells } from '@/utils/releaseScanCells'
 import Swal from 'sweetalert2'
 import { SwalData } from '@/utils/commonFunctions'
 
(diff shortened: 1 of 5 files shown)


taleodor-claude:
Verified live on psclaude with the Pro backend branch build and this UI build pinned:

| page | result |
|---|---|
| component (mafia-express, 16 branches) | Latest is the default tab; 16 rows newest first, each with a release link; rejected builds show the "Release rejected" status badge; clicking a row selects the branch and the right-hand pane follows |
| product (card-shuffle, 5 feature sets) | tabs Latest / Feature Sets; 5 rows with scan circles (0 / 11 / 15 …); row click selects the feature set |
| Pull Requests tab | gone; Branches tab shows PR-badged rows when `pull/…` branches exist (none in this org) |
| older backend (without the query) | tab renders "Could not load the latest releases (the backend may predate this view)" instead of breaking the page |

No GraphQL errors in any run. `latestReleasesOfComponentPerBranch` with the full UI selection: 16 rows in ~160 ms for the component.

PR #336, opened 2026-09-08
Author the per-release FDA narrative override, and read it in history (7h)

logicflakes:
Step 4b. The editor for the override rearm-saas#500 added, plus the history branch that
makes it readable. Small PR, three new files and one component.

## The org default is shown read-only beneath the editor

Without it, "override" is an instruction to write something without being told what it
displaces -- and the author cannot judge whether a device-specific justification is even
needed until they have read the one it would replace. When neither exists the panel says so
explicitly, because that combination means generated documents carry **no justification
section at all**.

## Same PATCH shape as the org prose form, including the lessons

* baseline diffed, so an unrelated release save cannot disturb the narrative
* `''` is a deliberate CLEAR, and the toast says what clearing **means**: the release returns
  to inheriting, it does not lose its justification
* the baseline is refreshed from the **mutation response**, before the refetch that can
  throw. On the org form, refreshing only via a nested re-read meant a failed re-read
  reported a COMMITTED save as failed and left the baseline stale, so the operator's next
  clear was silently omitted while the server kept the text
* the 8,000 cap is visible (`maxlength` + `show-count`), not only enforced server-side where
  an over-length paste fails the whole write and names the wire field
* `releaseNarrativeVariables` returns **null** for an unchanged form, so no mutation fires

## History: lengths, never the text -- and the length is the subtle part

`FDA_NARRATIVE` carries old/new values and no `objectId`, so without its own branch the row
falls through to `return row.objectId` and renders **blank** -- a Scope column saying the
justification changed beside nothing saying anything about it. Exactly the defect
`SUPPORT_WINDOW` had.

The backend stores a **200-character excerpt** in the event, appending `... (N characters)`
when it truncates. Using the stored value's own `.length` would tell an auditor that a
9,000-character justification was 218 characters -- a specific, plausible, wrong number in
the one place someone goes to reconstruct what happened. `narrativeLength` reads the marker
when present and falls back to the string only when the value was stored whole. Spec'd both
ways, including a marker-shaped phrase mid-sentence that must NOT be treated as the marker.

"cleared" is deliberately not "removed": clearing returns the release to INHERITING, so the
document keeps its justification section. An auditor reading "removed" would conclude the
opposite.

## Query pairing

Both release queries select the field, and `releaseFragmentsSchemaDrift` now asserts that
pairing for all three of `eos`/`eol`/`fdaAssessmentNarrative`. `fetchRelease` switches
documents once the artifacts tab is visited, and null here means INHERIT -- a meaningful
value, so an omission would be invisible rather than loud. That precise bug shipped once on
`eos`/`eol`.

The two single-release documents are now Pro-ahead, so the drift spec validates them against
Pro and **asserts the CE gap explicitly**: when the deferred sync lands that assertion fails,
and the documents move back to a CE check rather than the gap quietly becoming permanent.

## Pro-backend probe, all green

Against `rearm-pro:t3993-7g` on the sandbox, release `9000.1.0` (Infusion Pump 9000):

```
N1  narrative card present on the PRODUCT release
N2  org default shown read-only
N3  starts inheriting
N4  Save enabled once dirty
N5  override persisted across reload
N6  no longer says inheriting
N7  addendum carries the RELEASE override
N8  addendum scope says: this release
N9  FDA_NARRATIVE row present in history
N10 history row is NOT blank
N11 history does not leak the narrative text
N12 unrelated (device window) save left the narrative intact
N13 Save enabled for a clear
N14 override cleared
N15 says inheriting again
N16 addendum fell back to the ORG default
N17 addendum scope back to: organization default
X   0 page exceptions
```

N7/N8 and N16/N17 are the round trip through the resolution seam, observed rather than
inferred: set the override and the generated document changes what it carries and how it
labels the source; clear it and both go back.

## Checks

**586 tests / 43 files** (was 541/40). 0 non-ASCII added. 2 CE drift warnings, both expected
and both asserted so they cannot become permanent. `releaseNarrativeWiring.spec.ts` continues
the `*Wiring.spec.ts` pattern, and asserts the narrative is seeded on **every** path that
seeds the device window -- the window shipped with two load paths, and one missing seeding
call would blank the editor on the other.

### DIFF
--- a/ui/src/components/ReleaseView.vue
+++ b/ui/src/components/ReleaseView.vue
@@ -1070,6 +1070,51 @@
                             {{ deviceWindowError }}
                         </n-alert>
                     </div>
+                    <!-- The per-release FDA assessment narrative OVERRIDE. Beside the window
+                         because both are device-level facts an auditor reads together, and
+                         PRODUCT-gated for the same reason. The org default is shown
+                         read-only underneath so the author can see what they are replacing:
+                         without it, "override" is an instruction to write something without
+                         being told what it displaces. -->
+                    <div class="container" v-if="updatedRelease.componentDetails && updatedRelease.componentDetails.type === 'PRODUCT'">
+                        <h3>Assessment justification for this release</h3>
+                        <p class="text-muted" style="max-width: 760px;">
+                            Overrides the organization default below, for this release only.
+                            Leave it empty to inherit. Clearing a saved override returns this
+                            release to the default &mdash; it does not remove the
+                            justification from the generated documents.
+                        </p>
+                        <n-input v-model:value="releaseNarrative" type="textarea" :rows="5"
+                            style="max-width: 760px;"
+                            :maxlength="FDA_PROSE_MAX_LENGTH" show-count
+                            :disabled="!isWritable || savingReleaseNarrative"
+                            @update:value="releaseNarrativeError = null"
+                            placeholder="Leave empty to use the organization default." />
+                        <div style="margin-top: 8px;">
+                            <n-button v-if="isWritable" size="small" type="primary"
+                                :disabled="!releaseNarrativeDirty" :loading="savingReleaseNarrative"
+                                @click="saveReleaseNarrative">Save justification</n-button>
+                            <span v-if="!releaseNarrativeIsOverridden" class="text-muted"
+                                style="font-size: 12px; margin-left: 10px;">
+                                This release currently inherits the organization default.
+                            </span>
+                        </div>
+                        <n-alert v-if="releaseNarrativeError" type="error" :show-icon="true"
+                            style="font-size: 12px; max-width: 720px; margin-top: 10px;">
+                            {{ releaseNarrativeError }}
+                        </n-alert>
+                        <div style="margin-top: 14px;">
+                            <div class="text-muted" style="font-size: 12px;">
+                                Organization default (read-only, edited in Organization Settings)
+                            </div>
+                            <n-input v-if="orgNarrativeDefault" type="textarea" :rows="3"
+                                style="max-width: 760px;" readonly :value="orgNarrativeDefault" />
+                            <span v-else class="text-muted" style="font-size: 12px;">
+                                No organization default has been authored yet. With no override
+                                here either, generated documents carry no justification section.
+                            </span>
+                        </div>
+                    </div>
                     <div class="container" v-if="updatedRelease.componentDetails && updatedRelease.componentDetails.type === 'PRODUCT'">
                         <h3>Components
                             <Icon v-if="isWritable && isUpdatable"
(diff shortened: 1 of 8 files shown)


logicflakes:
## Both layers reported; findings fixed at `226e63b8`

### Layer 1 MEDIUM -- the success toast could state the opposite of what happened

`notify()` sat outside the `if (saved)` guard, and its wording keyed off
`releaseNarrativeBaseline` -- which, when the response is empty, is still the **pre-save**
value. A user with no override types 900 characters, clicks Save, the mutation resolves with
`data.updateRelease === null`: old baseline is `''`, so the toast reads *"Release
justification cleared; this release now inherits the organization default."* They just
authored a narrative and are told it was cleared. Clearing an existing one reports "updated".

Moved inside the guard, worded from the refreshed baseline, with the explicit else-warning
`OrgSettings` already uses. `saveDeviceWindow` shares the structural gap but its message is
direction-neutral, so this was the first place it produced an actively false statement.

### Layer 1 MEDIUM -- the drift assertion I wrote was near-worthless

`expect(errs.join(' ')).toContain('fdaAssessmentNarrative')` passes on a **broken** document:
an unrelated typo adds its own message while the expected string is still there. Combined
with the Pro check being skippable on a checkout without a `rearm-core` sibling, those two
documents could have ended up with no real validity checking -- and I had just moved them off
the CE check to make room for the expected drift. Every error must now be the expected one.
**Revert-probed**: injecting `artifactDetailsTYPO` into the product query fails it, where the
old form passed.

### Layer 1 LOW x2 -- the excerpt marker

Trusted unconditionally. The backend appends `... (N characters)` **only** above its
200-character cut, so a short narrative legitimately ending in a marker-shaped phrase (prose
quoting an earlier history row) was stored whole and reported 42 characters. Both conditions
now, revert-probed.

And nothing coupled that regex to the backend that writes it. A cross-repo assertion now
reads `ReleaseData.java` and checks both the cut and the format string -- the same trick the
schema drift specs use. Without it, a reword upstream would silently revert
`narrativeLength` to reporting ~218 for every long narrative, which is exactly the
wrong-number-in-the-audit-trail failure the file exists to prevent.

### Layer 2 -- two names for one thing, twice

Removed the `FDA_PROSE_MAX_LENGTH` re-export: the only `export { ... }` in `ui/src/utils`,
creating a second import path for one bound. Same objection that removed the analogous
constant alias in rearm-saas#500. And renamed the util predicate to
`releaseNarrativeDiffers` -- it was `releaseNarrativeDirty` and had to be aliased at the
import because the component's own computed owns that name, leaving one function tested under
one name and read under another.

---

## Layer 1's HIGH: correct observation, wrong conclusion -- and the wrong conclusion is my fault

It reported that adding the field to both single-release documents breaks the release page on
CE, and called it "a missed step, not an accepted deferral."

**It is a deferral** -- CE sync is deferred to feature-complete by operator ruling, drift is
informational, and no fallback is to be added. Layer 1 didn't know that: I put the ruling in
Layer 2's brief and left it out of Layer 1's. That's a prompt error on my part, not a bad
review.

**But it found something real inside it, and I have not seen this stated anywhere.** The
breakage is not confined to the new panel: both single-release documents now select the
field, so a CE backend rejects the **whole release query**. `store.fetchReleaseById` has no
drift tolerance, and `ReleaseView.onMounted` has no `try/finally` -- so `isLoading` and
`loadingBar` never clear. The failure mode is a **permanent spinner on every release page**,
product or component, not a missing section.

That is contained today (no CE images are being built per the ruling), but it changes the
shape of the sync: it is now a hard cutover. If the `Release`/`ReleaseInput` half is missed,
every release page hangs rather than degrading. #328 synced the **org-level** field only; the
per-release override landed later in #500.

Recorded on `t20260904-041253-3993`: the CE sync must add `fdaAssessmentNarrative` to CE
`Release` **and** `ReleaseInput` plus the CE `ReleaseData`/dto path, and the schema must gain
it **before** any CE image is built from this branch. Also noted the pre-existing missing
`try/finally` as worth closing separately, since it makes any release-query failure
unrecoverable without a reload.

**591 tests / 43 files.** 0 non-ASCII added. Probe unchanged (N1-N17 green, 0 page
exceptions) -- these fixes touch a toast, two assertions and two names.

PR #337, opened 2026-09-08
The FDA support addendum as PDF, sharing one document definition with the CSV (7i)

logicflakes:
Step 5. Extraction first, then the second renderer.

## addendumDocument.ts

The columns, the rows, the header block and the display order moved out of `addendumCsv.ts`.
Both renderers import from it and decide only how to **encode** what it says.

Those decisions are regulatory rather than typographic -- level and justification are
ALTERNATIVES under L1021-1022, a WITHDRAWN attestation states nothing at all, an absent fact
reads "not assessed" rather than leaving a blank a reviewer would count as an omission. A
module named "csv" exporting them to a PDF is a finding waiting to happen. Behaviour-neutral:
the 53 existing CSV and collector tests pass unchanged.

## Parity is structural, and the spec proves the wiring

The two documents cannot disagree about a cell because one function produces the rows.
`addendumParity.spec.ts` parses the CSV back with its own RFC-4180 reader and compares
against the PDF's table body: same columns, same order, same count, **same cell values**.

It earned its place immediately by finding the one real divergence: `csvCell` prefixes an
apostrophe on a cell starting with `=`/`+`/`-`/`@`, because spreadsheets evaluate those on
open and a justification rendering as a hyperlink means the exported evidence does not say
what the record says. A PDF has no such risk and carries the text verbatim. That is an
encoding difference, so the spec strips it and then asserts it is the **only** differing cell
in the document -- located by search, because `displayOrder` sorts alphabetically and the row
is not where it was declared.

**Revert-probed twice**: reimplementing the WITHDRAWN rule in the PDF alone fails "agree on
every cell value"; dropping the shared ordering fails "emit the rows in the same order".

## The renderer

`buildAddendumDocDefinition` returns **plain JSON and renders nothing**, so the spec asserts
structure directly instead of parsing a binary to learn whether a cell said "not assessed" --
which would test the parser as much as the document. 20 assertions, including the one that
matters most: no `noWrap`, no `ellipsis`, and a 9,000-character justification present in
full. A justification cut short at a page boundary is a regulatory statement the manufacturer
did not make, and it would look deliberate.

Not an extension of `pdfExport.ts`, deliberately: that module filters by severity and
analysis state, sorts by vulnerability type and colours by CVSS band. None of that means
anything here.

Footer on every page: release identity, the UTC RFC-3339 generation instant, page N of M.
Identity repeats because a page separated from the document must still say which release it
describes, and these get printed.

## A bug the unit specs could not have caught

`getBlob()` is promise-based in pdfmake 0.3 (`async getBlob()`). I passed a 0.2-style
callback; it was never invoked, so the promise never settled and **the export spinner span
forever with no error** -- the worst shape a failure can take, because it looks like
slowness.

The specs mock pdfmake, correctly, so the document can be asserted without rendering. That
leaves the pdfmake *seam* with no coverage, and that is precisely where the bug was. Two
assertions now close it: the promise must settle, and `getBlob` must be called with no
arguments. Revert-probed -- the callback form fails both by timeout, the same hang a user
would see.

## Pro probe, both encodings from one release

```
P4/P5   CSV data rows 10 == gauge total 10
P6      PDF is a real PDF
P7/P8   every CSV component and purl appears in the PDF
P9      10 of 10 rows carried
P9b     longest name intact, wrapped not cut
P10-P11 header facts present; PDF states the same total as the CSV
P12-P14 footer: page N of M, release uuid, UTC RFC-3339 instant
X       0 page exceptions
```

Two probe corrections worth recording, because **both made a correct document look broken**:

`pdftotext -layout` interleaves a wrapped cell's continuation with the other columns', so a
long purl is not contiguous even after stripping whitespace. Flowed mode reassembles cleanly.

Flowed mode then **de-hyphenates** -- it treats a hyphen at a line break as hyphenation and
drops it, and pdfmake wraps long identifiers at their hyphens, so `commons-collections` came
back as `commonscollections`. The comparison is hyphen-insensitive now, but it still reports
the strict count and separately proves from `-layout` that the hyphens are in the document.
Normalising a difference away without checking what it was is how a probe stops being
evidence.

## Checks

**620 tests / 46 files** (was 591/43). 0 non-ASCII added. Build clean. 2 CE drift warnings,
unchanged and expected.

### DIFF
--- a/ui/src/components/ReleaseView.vue
+++ b/ui/src/components/ReleaseView.vue
@@ -85,7 +85,7 @@
                          addendum always walks the whole release scope, so an operator who
                          ticked "Top Level Dependencies Only" and got a full-scope document
                          would have no way to tell the control had been dropped on the floor. -->
-                    <n-form-item v-if="selectedSbomMediaType !== 'FDA_ADDENDUM'">
+                    <n-form-item v-if="!isAddendumExport">
                         <template #label>
                             <span style="display: inline-flex; align-items: center;">
                                 Select SBOM configuration for export
@@ -123,9 +123,10 @@
                                  below do not apply to it. A checkbox would imply it rides
                                  along with whichever format was picked. -->
                             <n-radio-button value="FDA_ADDENDUM">FDA support addendum (CSV)</n-radio-button>
+                            <n-radio-button value="FDA_ADDENDUM_PDF">FDA support addendum (PDF)</n-radio-button>
                         </n-radio-group>
                     </n-form-item>
-                    <n-alert v-if="selectedSbomMediaType === 'FDA_ADDENDUM'" type="default"
+                    <n-alert v-if="isAddendumExport" type="default"
                         :show-icon="false" style="font-size: 12px; max-width: 620px; margin-bottom: 10px;">
                         Every component in this release with its level of support, the end-of-support
                         date where one is attested and the justification where none is, plus the
@@ -157,7 +158,7 @@
                             />
                         </n-radio-group>
                     </n-form-item>
-                    <n-form-item v-if="selectedSbomMediaType !== 'FDA_ADDENDUM'">
+                    <n-form-item v-if="!isAddendumExport">
                         <span style="display: inline-flex; align-items: center;">
                             Top Level Dependencies Only:
                             <n-tooltip trigger="hover">
@@ -171,7 +172,7 @@
                         </span>
                         <n-switch style="margin-left: 5px;" v-model:value="tldOnly"/>
                     </n-form-item>
-                    <n-form-item v-if="selectedSbomMediaType !== 'FDA_ADDENDUM'">
+                    <n-form-item v-if="!isAddendumExport">
                         <span style="display: inline-flex; align-items: center;">
                             Ignore Optional Dependencies:
                             <n-tooltip trigger="hover">
@@ -1750,6 +1751,7 @@ import { GET_VEX_PROPOSALS_BY_RELEASE } from '@/graphql/vexImport'
 import commonFunctions, { SwalData } from '@/utils/commonFunctions'
 import { collectAddendumData } from '@/utils/addendumData'
 import { renderAddendumCsv, addendumFileName } from '@/utils/addendumCsv'
+import { renderAddendumPdfBlob, addendumPdfFileName } from '@/utils/addendumPdf'
 import { releaseNarrativeVariables, releaseNarrativeDiffers } from '@/utils/releaseNarrativeInput'
 import { FDA_PROSE_MAX_LENGTH } from '@/utils/fdaProseInput'
 import { formatNarrativeChange } from '@/utils/narrativeHistory'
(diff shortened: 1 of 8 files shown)


logicflakes:
## Both layers reported; all findings fixed at `fdca951b`

### Layer 1 HIGH -- non-Latin text vanished from the PDF

pdfmake bundles **Roboto only**, and a character it has no glyph for is not flagged,
substituted or warned about -- it is silently **dropped**. A component named in Japanese
produced an empty Component cell, and a Chinese justification an empty Justification cell, in
a regulatory document -- while the CSV for the same release showed the text (`csv.ts` goes
out of its way to add a BOM for exactly this).

Blank is the one thing this document must never be ambiguous about. Its own doc comment says
an absent fact reads "not assessed" *precisely so* a blank cannot mean two things, and this
reintroduced a blank meaning neither.

Layer 1 rendered the document to prove it rather than inferring, and I verified independently
before acting -- rendered through pdfmake 0.3.11, extracted with `pdftotext` -- then mapped
coverage the same way instead of reading a spec:

| script | result |
|---|---|
| Latin incl. Extended-A, Turkish, Vietnamese | draws |
| Greek, Cyrillic, punctuation, currency | draws |
| Latin Extended-B, arrows, emoji | dropped |
| Hebrew, Arabic, Hangul, Thai, Devanagari, CJK | dropped |

`findUnrenderableText` scans **every cell** -- header block and component rows, since the
first report was a component *name* -- and the export refuses in the same voice as every
other addendum refusal: no document, a reason, the offending characters named, and a pointer
to the CSV which carries them correctly.

The range list is a **whitelist and deliberately conservative**: an unlisted script is
refused rather than rendered blank, so it will refuse some characters Roboto could in fact
draw. That trade is the point -- a false refusal is visible, explained and recoverable; a
false render is an invisible hole in a document someone files. Revert-probed: disabling the
refusal fails 10 assertions.

### Layer 1 MEDIUM x3

**The title was a literal in three places and the de-dup was a string match** -- rename it and
the PDF printed it twice, once as a heading and once as an empty fact row, while the guard
test *stayed green* because it asserted the absence of the old literal. `ADDENDUM_TITLE` is
owned by `addendumDocument.ts` now. Revert-probed by re-introducing the copy.

**`addendumPdfFileName` re-implemented the CSV's slug rule verbatim**, and both specs
hardcoded their expected filenames -- so changing the rule in one would pass every test while
breaking the "shares the CSV stem" contract. Shared as `addendumFileStem`, and it no longer
throws when version and uuid are both absent.

**One BOM-shaping control escaped the gate.** The artifact-coverage-type filter had no
`v-if`, so it stayed visible and toggleable for both addendum encodings while the handler
ignored it entirely -- the exact "dropped on the floor" failure this PR's own comment
describes for `tldOnly`. Pre-existing, but I re-gated its three siblings and left it. The
spec asserted `>= 3`, which could never catch it; it asserts exactly 4 now.

### Layer 2 -- I argued against a two-value test and then wrote one

`asPdf: boolean` in three places, while the comment justifying `isAddendumExport` says adding
a third encoding "cannot leave one control wired to the old two-value test", and the
neighbouring `exportReleaseSbom` already carries format as a value. The handler takes
`mediaType` now.

Also from both layers: derived `colSpan` filler instead of seven hardcoded cells;
`displayName` private again (the extraction widened it with no caller); `cell()`'s doc comment
corrected, since it claimed a cell "must never be blank" directly above a branch returning
`''`; a spinner regression guard; and four parity-spec gaps closed -- an unguarded index that
turned a row-count divergence into a `TypeError`, a one-directional header comparison that
would have missed a dropped fact, the unasserted empty-release case, and the tab divergence
named rather than left to surprise someone.

### Re-probed on Pro after the fixes -- unchanged and green

```
P4/P5   CSV data rows 10 == gauge total 10
P6      PDF is a real PDF
P7/P8   every CSV component and purl appears in the PDF
P9/P9b  10 of 10 rows; longest name intact, wrapped not cut
P10-P11 header facts present; PDF states the same total as the CSV
P12-P14 footer: page N of M, release uuid, UTC RFC-3339 instant
X       0 page exceptions
```

**643 tests / 46 files** (was 620). 0 non-ASCII added -- the non-Latin fixtures are `\u`
escapes, since the CI invisible-character check runs before the build.

PR #338, opened 2026-09-08
The Device Support Statement, third renderer over one collector (7j)

logicflakes:
Step 6. FDA labeling section VI.A, customer-facing, PDF only.

## Written for a different reader

L1591-1592 says the audience "might include patients or caregivers with limited technical
knowledge". That single constraint decides most of the document: **no purls, no component
inventory, no version column, no unassessed count, no `deviceSupportRisk` verdict.**

The unassessed count is excluded **deliberately, not by omission**. "412 components not
assessed" conveys no risk information to a hospital biomed reader and edges toward misleading
under 502(a)(1); it belongs in the addendum, which a reviewer reads. A spec asserts it cannot
appear.

## It says on its face what it does not cover

VI.A lists four asks. This answers **one** -- support dates -- and names the other three as
commitments the manufacturer addresses separately, so nobody reads shipping it as satisfying
VI.A. Under section 3 there is no product-shipped default text for any of the three, which is
exactly why they are org-authored prose rather than generated.

## Blocked, never degraded

Three refusals, all in `statementBlockReason` so a caller cannot check some and forget
others, ordered so the most fundamental problem is the one reported:

1. a **COMPONENT release is not a device**, and no amount of authored prose makes it one;
2. an **unauthored prose slot** blocks with the slot named and Organization Settings pointed
   at. Section 3's "a gap beats a fabrication" is right for a submission where a reviewer
   expects gaps; on a patient-facing document a silent gap is itself the misleading thing;
3. **text the PDF font cannot draw**, reused from the addendum.

## The data slots

Device EOS and EOL are two separate labelled facts, "not declared" when null, and
deliberately **not** routed through a support-horizon fallback -- EOS-else-EOL is right for a
risk comparison and wrong for a document read as two distinct commitments.

Components appear only where an **attested** EOS precedes the device EOS, by name and date,
as recorded facts. Not listed is **not** a claim that something outlives the device -- it
means nobody recorded a date, and asserting anything about it to a patient would be a
fabrication. With no device EOS there is nothing to precede, so the list is empty rather than
"every dated component", which would silently change what the section means.

## One collector change, flagged rather than buried

The prose slots were all already collected, as expected. But the release query fetched
`componentDetails { name type }` and the collector kept only the name -- so `componentType`
is kept now. One field the query already fetches; the alternative (telling the renderer its
own release's type from the call site) would be a second source of truth for something the
collector already holds. Happy to invert it if you'd rather the collector stayed frozen.

The font check moved to `pdfFontCoverage.ts`, shared by both PDFs: the failure it prevents is
a property of the **font**, not of one document.

## Pro probe, all four cases

```
S1-S9  product release, all slots set -> generates
       EOS 2031-01-31 and EOL "not declared" as two separate facts
       one component listed: openssl, attested EOS 2027-06-30
       no purl, no unassessed count, VI.A one-of-four stated on its face
       footer: device identity, UTC instant, page N of M
S10    component release -> REFUSED, told to open the product release
S11    cleared prose slot -> BLOCKED, slot named, Organization Settings pointed at
       restoring the text generates again
S12    CJK in a prose slot -> FONT refusal naming the characters
       restoring the text generates again
X      0 page exceptions
```

The generated document, verbatim:

```
Device software support statement
Infusion Pump 9000 (device), version 9000.1.0

What this document covers
This document provides end-of-support and end-of-life information for this device and its
software, which is one of the four items FDA labeling guidance section VI.A suggests may be
included in labeling.
It does not cover the following, which the manufacturer addresses separately:
  That the manufacturer may no longer be able to provide security patches...
  The manufacturer's pre-established process for transferring cybersecurity risk...
  That cybersecurity risk to users can be expected to increase over time.

Support dates for this device
Software support ends (end of support)     2031-01-31
Device end of life / end of sale           not declared

Software components whose support ends sooner
openssl                                    support ends 2027-06-30
```

## Out of scope, filed as `t20260908-053642-13923`

Delivery as a **versioned shared-artifact link**. VI.A (L1592-1594, L1636-1640) requires
portal links to stay up to date, and a downloaded snapshot goes stale the moment a device
window is revised. D1-independent by the plan's own reasoning, so it does not block this
document -- but this PR ships a download, not a link.

## Carry-over

`addendumDocument`'s tests moved out of `addendumCsv.spec.ts` into their own spec, so the
module boundary #337 drew is reflected in the specs too.

**694 tests / 49 files** (was 643/46). 0 non-ASCII added -- the CJK fixtures are `\u` escapes.

### DIFF
--- a/ui/src/components/ReleaseView.vue
+++ b/ui/src/components/ReleaseView.vue
@@ -85,7 +85,7 @@
                          addendum always walks the whole release scope, so an operator who
                          ticked "Top Level Dependencies Only" and got a full-scope document
                          would have no way to tell the control had been dropped on the floor. -->
-                    <n-form-item v-if="!isAddendumExport">
+                    <n-form-item v-if="!isFdaDocumentExport">
                         <template #label>
                             <span style="display: inline-flex; align-items: center;">
                                 Select SBOM configuration for export
@@ -124,6 +124,11 @@
                                  along with whichever format was picked. -->
                             <n-radio-button value="FDA_ADDENDUM">FDA support addendum (CSV)</n-radio-button>
                             <n-radio-button value="FDA_ADDENDUM_PDF">FDA support addendum (PDF)</n-radio-button>
+                            <!-- A different DOCUMENT for a different reader: FDA L1591-1592
+                                 says the audience may include patients or caregivers with
+                                 limited technical knowledge, so it carries no component
+                                 inventory and no counts. PRODUCT releases only. -->
+                            <n-radio-button value="DEVICE_STATEMENT">Device support statement (PDF)</n-radio-button>
                         </n-radio-group>
                     </n-form-item>
                     <n-alert v-if="isAddendumExport" type="default"
@@ -158,7 +163,7 @@
                             />
                         </n-radio-group>
                     </n-form-item>
-                    <n-form-item v-if="!isAddendumExport">
+                    <n-form-item v-if="!isFdaDocumentExport">
                         <span style="display: inline-flex; align-items: center;">
                             Top Level Dependencies Only:
                             <n-tooltip trigger="hover">
@@ -172,7 +177,7 @@
                         </span>
                         <n-switch style="margin-left: 5px;" v-model:value="tldOnly"/>
                     </n-form-item>
-                    <n-form-item v-if="!isAddendumExport">
+                    <n-form-item v-if="!isFdaDocumentExport">
                         <span style="display: inline-flex; align-items: center;">
                             Ignore Optional Dependencies:
                             <n-tooltip trigger="hover">
@@ -193,7 +198,7 @@
                         </span>
                         <n-switch style="margin-left: 5px;" v-model:value="ignoreDev"/>
                     </n-form-item>
-                    <n-form-item v-if="!isAddendumExport">
+                    <n-form-item v-if="!isFdaDocumentExport">
                         <div style="width: 100%;">
                             <div style="display: inline-flex; align-items: center;">
                                 <span style="display: inline-flex; align-items: center;">
@@ -1752,6 +1757,8 @@ import commonFunctions, { SwalData } from '@/utils/commonFunctions'
 import { collectAddendumData } from '@/utils/addendumData'
 import { renderAddendumCsv, addendumFileName } from '@/utils/addendumCsv'
 import { renderAddendumPdfBlob, addendumPdfFileName, findUnrenderableText } from '@/utils/addendumPdf'
+import { renderDeviceSupportStatementBlob, deviceSupportStatementFileName,
+    statementBlockReason } from '@/utils/deviceSupportStatement'
 import { releaseNarrativeVariables, releaseNarrativeDiffers } from '@/utils/releaseNarrativeInput'
 import { FDA_PROSE_MAX_LENGTH } from '@/utils/fdaProseInput'
 import { formatNarrativeChange } from '@/utils/narrativeHistory'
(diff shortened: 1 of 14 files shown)


logicflakes:
## Both layers reported; all findings fixed at `009ee2db`

### Layer 1 HIGH -- the document declared it did not cover what it then printed

The generated PDF said *"It does not cover the following, which the manufacturer addresses
separately:"* and listed patches-may-cease, the risk-transfer process and risk-increases --
then printed all three, in the same order, under their own headings. And
`REQUIRED_PROSE_SLOTS` **blocks generation unless all three are authored**, so the document
was refused unless it covered exactly what it claimed not to. Every statement it produced
contradicted itself on its face: the 502(a)(1) misleading-labeling risk this module's own
header invokes.

The root cause was mine -- I flattened the plan's distinction between what is **generated
from records** (the dates, answering one of VI.A's four items) and what is **reproduced
verbatim** (the three manufacturer statements, neither generated nor checked here). The
risk-transfer item keeps its own sentence because there the gap is substantive rather than
editorial: the slot holds a *reference* to a controlled DHF document, so the process itself
genuinely is not in this document.

**Both specs that were supposed to guard that block asserted the wrong text and stayed green
through it.** They now assert the invariant: no "does not cover" phrasing anywhere, and each
prose slot appearing exactly once.

The corrected block, from the live PDF:

```
What this document covers
The support dates below are generated from the manufacturer's records. They answer one of
the four items FDA labeling guidance section VI.A suggests may be included in labeling, and
it is the only one of the four this document generates.

The three statements that follow the dates are the manufacturer's own words, reproduced here
unchanged. They are not generated, completed or verified by the system that produced this
document.

Where this document refers to the manufacturer's process for transferring risk, it gives a
reference to that controlled document. The process itself is held separately and is not
reproduced here.
```

### Layer 1 MEDIUM x3

**One library at three versions rendered three rows.** The version is deliberately hidden (it
means nothing to this reader), so three attested versions of `log4j-core` produced three rows
saying different things about the same name. Collapsed by name to the **earliest** date --
the one that matters here, being when this reader first loses supported software.

**`renderDeviceSupportStatementBlob` did not enforce its own precondition.** The only guard
was in the caller; the builder casts the prose slots to string and would emit `{ text: null }`
into pdfmake -- a silent empty section, exactly what the block exists to prevent. It is
exported and nothing type-checks the template that calls it, so one guard in one caller was
not a guarantee. It throws now.

**`statementStrings` claimed to be "every string the document prints"** and omitted the device
dates, the release uuid and the generation timestamp, all of which are rendered. Unreachable
today, but the next field added to the footer would have escaped the font whitelist silently.

### Layer 2 -- two things I should have caught myself

**I re-duplicated the slug rule #337 had just de-duplicated**, character for character,
including both specs hardcoding their expected filenames independently -- reproducing both
halves of the regression whose docstring exists to explain why it was shared. `releaseSlug`
is now shared by all three documents.

**And for the fourth time on this track I spliced a function between an existing docblock and
the function it documents**, leaving `exportFdaAddendum` undocumented with two blocks stacked
on the new one. Same mistake as `ReleaseDto`, `ReleaseData` and `schema.graphqls` in #500.

Also: the three prose-slot labels are shared from `addendumDocument` rather than copied, so a
block message and an addendum header cannot name the same field differently; and the modal
now states the statement is PRODUCT-only *before* an export attempt walks the whole component
list to refuse.

### Re-probed after the rewrite, since the document text changed materially

```
S1-S9   generates; EOS and EOL separate; one component listed
S3      states it answers one of VI.A's four
S4/S4b  distinguishes generated from reproduced; says the process is not reproduced here
S4c     NEW -- never claims not to cover what it prints
S5/S6   no purl, no unassessed count
S10     component release refused
S11     cleared slot blocked, slot named, settings pointed at; restoring generates again
S12     CJK in a slot hits the font refusal; restoring generates again
X       0 page exceptions
```

One note on the probe run: S11/S12 clear and rewrite org settings, and an earlier attempt
died mid-way while the sandbox backend was restarting under memory pressure -- leaving
`fdaPatchesMayCeaseStatement` cleared. I checked `kubectl get pods` before trusting the
failure, restored the slot, verified all three in the database, and re-ran clean. The
sandbox is back to its seeded state.

**702 tests / 48 files.** 0 non-ASCII added. Build clean.

PR #339, opened 2026-09-08
The export injection toggle, and the state stated beside the gauge (7g A-ui)

logicflakes:
7g PR A, UI half. Pairs with rearm-saas#504 (merged), which added the setting and the single
predicate this reads.

## The state is now said out loud

Decision D3 calls this a build requirement, not a suggestion. With injection defaulting OFF,
a gauge reporting full attestation coverage beside an export carrying nothing is a lie by
omission -- so the coverage line now reads
`N of M components have a support disclosure -- export injection ON/OFF`, rather than leaving
the export state to be inferred from whether a warning happens to be showing.

## All three states, and UNKNOWN is not OFF

`DISABLED` says the organization chose not to export attestations. `UNKNOWN` says the setting
could not be read. Rendering the second as the first states a choice nobody made, so it reads
**"export injection could not be read"** and the operator is told to find out rather than
reassured. A value this build does not recognise -- a newer server introducing a state --
gets the same wording, never OFF, for the same reason.

`stateLabel` is null in exactly two places: before the first load resolves, and after a failed
one. There is no state to name there, and naming one would be a claim built on a request that
did not return.

## The toggle

Sits with the FDA prose in Organization Settings, because it is the same decision -- what a
generated document or export says about support -- and because that is where a manufacturer
preparing a submission is already working.

**A Boolean in the form, the enum on the wire.** The schema deliberately uses a two-member
enum so the field can grow a third state without breaking published clients; a switch is what
an operator expects for a two-state choice today. The mapping lives in one place so the form
cannot invent a value the schema does not declare, and a wiring spec asserts it never sends a
bare boolean.

Only `ENABLED` is on -- `DISABLED`, null, unset and unrecognised all read as off, matching
the server's own default rule. The baseline is refreshed from the mutation's own response
before anything that can throw, the same rule the prose baseline follows: a failed re-read
must not leave the form believing an unsaved state was saved. Sent only when changed, since
the mutation is a PATCH.

The switch copy says plainly that the strip is unaffected: *"Forged support properties are
always removed, whatever this is set to; this controls only whether our own attestations are
added."* Someone turning injection off must not believe they have turned off a security
control.

`PARTIAL`'s gauge copy is kept and marked retired. The server no longer returns it, but a UI
build can meet an older server, and that sentence is more accurate for a genuinely split
server than the unknown fallback would be.

## Checks

**719 tests / 51 files** (was 708). New: 6 for the three states beside the gauge, 7 wiring
assertions for the toggle, and a drift spec pinning the setting against the Pro schema --
including that the enum has exactly the two members the switch maps to, so a third member
tells whoever adds it that a switch is no longer sufficient.

Build clean. 0 non-ASCII added. 4 CE drift warnings, up from 2: the two new `supportInjection`
selections, expected and asserted, since CE gains the field at the deferred sync.

**Note for review: build + vitest is not the gate here.** There is no vue-tsc in this repo, so
an undefined identifier in `<script setup>` is a runtime error the build ignores -- this
feature has already shipped a blank panel that way once. The Layer 1 pass on this PR includes
a browser probe of all three states.

### DIFF
--- a/ui/src/components/OrgSettings.vue
+++ b/ui/src/components/OrgSettings.vue
@@ -994,6 +994,26 @@
                             </div>
                         </n-form-item>
 
+                        <!-- The export toggle sits WITH the FDA text because it is the same
+                             decision: what a generated document or export says about support.
+                             Default OFF (D3), so this is the control a manufacturer preparing
+                             a submission must deliberately turn on. -->
+                        <n-form-item label="Carry support attestations in BOM exports">
+                            <div style="display: flex; flex-direction: column; width: 100%;">
+                                <n-switch v-model:value="supportInjectionEnabled"
+                                    :disabled="savingOrgSettings" />
+                                <span class="text-muted" style="margin-top: 4px; max-width: 760px;">
+                                    Off by default. When on, BOM exports carry this
+                                    organization's support attestations &mdash; the artifact
+                                    download, the SPDX-augmented download and the release SBOM
+                                    export. The raw artifact download never carries them.
+                                    <strong>Forged support properties are always removed,
+                                    whatever this is set to</strong>; this controls only
+                                    whether our own attestations are added.
+                                </span>
+                            </div>
+                        </n-form-item>
+
                         <n-form-item label="Risk increases over time (labeling)">
                             <n-input v-model:value="orgSettings.fdaRiskIncreasesNotice"
                                     :disabled="savingOrgSettings" :maxlength="FDA_PROSE_MAX_LENGTH" show-count
@@ -3516,6 +3536,17 @@ async function saveOrgDefaultView() {
  */
 const proseBaseline: Record<string, string> = {}
 
+/**
+ * The export toggle as a Boolean for n-switch, mapped to the two-member enum on save.
+ *
+ * A Boolean in the FORM and an enum on the WIRE: the schema deliberately uses an enum so the
+ * field can grow a third state without breaking published clients, but there are two states
+ * to choose between today and a switch is what an operator expects for that. The mapping
+ * lives in one place, here, so the form cannot invent a value the schema does not declare.
+ */
+const supportInjectionEnabled: Ref<boolean> = ref(false)
+const supportInjectionBaseline: Ref<boolean> = ref(false)
+
 async function loadOrgSettings() {
     await loadOrgDefaultView()
     const s = myorg.value?.settings
@@ -3527,6 +3558,10 @@ async function loadOrgSettings() {
     orgSettings.sidAuthoritySegments = Array.isArray(s?.sidAuthoritySegments)
         ? [...s.sidAuthoritySegments]
         : []
+    // ENABLED is the only truthy value; anything else -- DISABLED, null, unset, or a state
+    // this build does not know -- reads as off, which matches the server's own default rule.
+    supportInjectionEnabled.value = s?.supportInjection === 'ENABLED'
+    supportInjectionBaseline.value = supportInjectionEnabled.value
     const seeded = proseBaselineFrom(s)
     for (const f of FDA_PROSE_FIELDS) {
         orgSettings[f] = seeded[f]
@@ -3573,6 +3608,7 @@ async function saveOrgSettings() {
                             fdaPatchesMayCeaseStatement
                             fdaRiskTransferProcessRef
                             fdaRiskIncreasesNotice
+                            supportInjection
                         }
                     }
                 }`,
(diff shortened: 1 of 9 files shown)


logicflakes:
## Browser-rendered against the real backend. The toggle has been seen.

Sandbox rebuilt from merged `fda-readiness-1-dev` (`01331db4`) as `rearm-pro:t10180-504`,
hot-swapped under the `cluster` lane; `kubectl get pods` confirmed before every reading. UI
served from this branch's head (`529373a4`).

### 1. The toggle renders, saves both ways, and leaves the prose alone

```
T1 page renders (not blank), body chars: 2979
T2 TOGGLE IS VISIBLE: true
T3 initial state: OFF
T4 after save #1 -> ON:  true
T5 after save #2 -> OFF: true
T6 after save #3 -> ON:  true
T7 three saves observed: true | any GraphQL errors: none
X  page exceptions: 0
```

Prose checked against the **server's own response** to each save, not against the form. All
four slots identical across all three saves:

```
fdaAssessmentNarrative="narrative 1788784329455" | fdaPatchesMayCeaseStatement="patches statement seed"
| fdaRiskTransferProcessRef="DHF-SEED-2 rev B" | fdaRiskIncreasesNotice="risk notice seed"
```

Three saves in both directions, so it is not a one-way latch.
Screenshots: `shot-toggle-visible.png`, `shot-toggle-on.png`.

### 2. The gauge, all four states

| state | rendered line |
|---|---|
| ENABLED | `1 of 10 components have a support disclosure. — export injection ON` |
| DISABLED | `... — export injection OFF` |
| UNKNOWN | `... — export injection could not be read` + *"Export state could not be determined — do not assume exports carry the disclosure."* |
| PARTIAL | `... — export injection PARTIAL -- do not assume the release export carries it` |

ENABLED and DISABLED were driven **through the real toggle** end to end. UNKNOWN and PARTIAL
were forced on the wire, and here is why that is the honest method rather than a shortcut: a
bad org uuid does **not** produce UNKNOWN. `sbomComponentSupportCoverage` authorizes the org
first, so an unreadable org fails authorization and lands in the display's ERROR branch --
correctly a different thing. UNKNOWN is the server saying "I could not determine the setting"
for an org you *are* allowed to see.

`never says OFF` asserted in both non-determinate states. 0 page exceptions throughout.
Screenshots: `shot-gauge-{enabled,disabled,unknown,partial}.png`.

### 3. The fallback, against a backend WITHOUT the field

This is the one that matters, since it is the HIGH. Same bundle, old backend
(`rearm-pro:t3993-7g`):

```
F1 PAGE IS NOT BLANK, body chars: 2595      (it was 0 before the fix)
F2 nav present: true
F3 FDA prose section still rendered: true
F4 TOGGLE HIDDEN (server cannot store it): true
F5 release page renders, gauge line present
X  page exceptions: 0 | graphql >=400 responses: 7
```

The 7 GraphQL 400s are the drift fallback and the pre-existing `OrgDefaultView` failure doing
their job: the queries that cannot be answered fail, the app renders anyway, and the control
the server cannot back is hidden rather than offered.
Screenshots: `shot-fallback-orgsettings.png`, `shot-fallback-release.png`.

### State of the sandbox

Backend restored to `rearm-pro:t10180-504`, cluster lane released, `supportInjection` left ON
from the gauge run -- the ingest probe sets it explicitly and restores DISABLED in a `finally`.

**726 tests / 53 files**, build clean, 0 non-ASCII, CE drift back to 2.

logicflakes:
## Screenshots recaptured — the first set did not show what I said it did

You were right on both counts, and the first duplicate was worse than a duplicate:
`shot-toggle-on.png` and `shot-fallback-orgsettings.png` were taken against **different
backends** and were byte-identical. That alone proves the toggle was never in frame — one
should show it and the other should not. I captured the top of OrgSettings both times and
asserted things about a region the image did not contain.

### Recaptured, element-scoped

**(a) Toggle ON, new backend (`rearm-pro:t10180-504`), after a save and a reload.** Captured
as an element screenshot of the form-item itself, so it cannot be out of frame:

```
toggle form-item present: true
state after save:        true        (aria-checked, re-read after reload)
label text in frame:     Carry support attestations in BOM exports
```

The image shows the label, the switch green/on, and the help text including *"Forged support
properties are always removed, whatever this is set to"*.

**(b) Same region, old backend (`rearm-pro:t3993-7g`), toggle absent.** The toggle sits
between "Risk-transfer process reference" and "Risk increases over time" when present, so the
capture is a band anchored on those neighbours — the absence is **visible**, not merely
asserted: the two prose fields are adjacent with nothing between them, running straight on to
"VEX Compliance Framework".

```
toggle form-item present: false
```

### md5sum, after recapture

```
88509d7362fd5b32a8454db9cd9e667d  shot-fallback-orgsettings.png
985842d222927bdda7ed7d005af991e6  shot-gauge-disabled.png
ad273bb9261b353ead6e37cf409c5c04  shot-gauge-enabled.png
1cb435606952abdbea76711a6a17f347  shot-gauge-partial.png
bf1b0173dfafb6558fca8289e9d853be  shot-gauge-unknown.png
ff901f8680db3bfec751d161fbd1c7c9  shot-toggle-on.png
c1a3a27ed921d2545a5bef7d43f48765  shot-toggle-visible.png
```

All seven distinct. `shot-fallback-release.png` is deleted rather than listed: the old backend
genuinely returns PARTIAL, so the fallback release page and the forced-PARTIAL page are the
same rendered state, and `shot-gauge-partial.png` is that one image.

### Sandbox

Backend restored to `rearm-pro:t10180-504`, cluster lane released, `supportInjection` left ON
from the recapture — the ingest probe sets it explicitly and restores DISABLED in a `finally`.

PR #341, opened 2026-09-08
Document what support attestations put in an export (7g PR B docs)

logicflakes:
Documents what rearm-saas #505 (7g PR B) makes visible in a served SBOM. Companion to
that PR; no code.

## Why

7g changes the export shape a consumer sees -- the `reliza:support:*` properties, the
new document-level `declarations` block, and the disclosure marker saying which of the
two things happened. None of it was documented anywhere.

## Placement -- flagging a deviation

The instruction was `documentation_site/docs/integrations/`. I put it in
`workflows/exporting-compliance-artifacts.md` instead.

`integrations/` is entirely third-party connectors -- Slack, Jenkins, Dtrack, BEAR, ADO,
GitHub -- and follows the rule that every outbound third-party call needs a page there.
An export shape is not an outbound call, and someone asking "what is in the SBOM ReARM
just gave me" opens the export page, which already documents SBOM/VDR/VEX/OBOM/BOV.
**Say the word and it moves.**

## What it covers

The section a reader most needs is the disclosure marker, so it is called out rather
than tabulated away:

- `derived-non-attested-current-state` -- an absent support property means no
  attestation on file.
- `provenance-stripped-no-disclosure` -- **nothing was asserted**, including about
  components that DO have a recorded end-of-support date.

Reading the second as the first is how a reviewer concludes a device has no
out-of-support parts when it does.

Also states two things the UI cannot:

- The setting is **off by default**, so full coverage on the gauge beside an export
  carrying none of it is the default rather than a contradiction.
- The reliza namespaces and the `declarations` block are server-owned and stripped from
  uploads on every egress **whether or not the setting is on**, because a security
  control and a content choice are not the same switch.

Plus the property table, the `declarations` shape and its 1.6 floor, and the note that
CSV/Excel media types carry no component properties at all.

### DIFF
--- a/documentation_site/docs/workflows/exporting-compliance-artifacts.md
+++ b/documentation_site/docs/workflows/exporting-compliance-artifacts.md
@@ -67,12 +67,24 @@ Two layers, both emitted together and both governed by that single setting.
 | `reliza:support:source:<milestone>` | Where each milestone date came from |
 | `cdx:lifecycle:milestone:endOfSupport` (and `endOfLife`, `endOfGuaranteedSupport`) | Standard CycloneDX milestone dates |
 
-**A `declarations` block** at the document level, in CycloneDX's own attestation vocabulary --
-`claims[]` naming what was attested about which component, `evidence[]` carrying the value, the
-assessment instant and the person who recorded it, and `assessors[]` naming the organization.
+**A `declarations` block** at the document level, in CycloneDX's own attestation vocabulary:
+
+- `claims[]` -- what was attested about which component. A claim targets a component's
+  `bom-ref`; where a component arrived without one, ReARM assigns a namespaced
+  `reliza:bomref:` identifier so the claim has something to point at, and removes it again on
+  any export that does not carry the claim.
+- `evidence[]` -- the attested value, the assessment instant, and the person who recorded it.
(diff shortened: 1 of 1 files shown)


logicflakes:
Updated in `399e83db` to match what rearm-saas#505 actually ships after its review pass:

- **Milestone dates are evidence too**, each with its own assessment instant -- the review found Layer B carrying no dates at all, which is the fact a reviewer of an FDA submission actually came for.
- **`attestations[]`** joins an assessor to the claims it made; previously nothing linked them.
- **`reliza:bomref:`** -- the namespaced identifier ReARM assigns to a component that arrived without a `bom-ref`, and removes again on exports that do not carry the claim.
- **The value agreement is now stated explicitly.** `reliza:support:party` reads `FIRST_PARTY`/`THIRD_PARTY` in BOTH layers; the review found it emitting the CycloneDX role in one and the enum constant in the other, which would have made the reconciliation this page promises impossible. The role appears only on the assessor.

Placement deviation still stands for your call: this is in `workflows/exporting-compliance-artifacts.md` rather than `documentation_site/docs/integrations/`, because that directory is third-party connectors and an export shape is not an integration.

PR #343, opened 2026-09-09
Walkthrough findings (UI): stale toggle, toggle placement, statement gate, withdrawn presentation

logicflakes:
UI half of the walkthrough findings (board `t20260909-061338-23148`). Backend companion:
relizaio/rearm-saas#509. **Commits are in the order the findings were filed.**

| Commit | Finding |
|---|---|
| `b2494a3b` | **(a) HIGH** -- the support-injection toggle went stale in the store after a save |
| `96b2d900` | **(b)** -- the toggle rendered inside the labeling prose slots |
| `6332f1ca` | **(c)** -- statement Export stayed enabled under the PRODUCT gate |
| `3f3d6ff3` | **(d)** -- withdrawn chip, and why Save is disabled |
| `123c8c03` | **(f)** -- the statement's ends-sooner section is a sentence, not an inventory |

## (a) HIGH -- the toggle lied about the disclosure state

`updateOrganizationSettings` deliberately does not select `supportInjection`: adding it makes
the whole mutation invalid against a CE backend without the field, failing every save including
the prose slots. But `UPDATE_ORGANIZATION` **replaces** the stored organization, so committing
that response dropped the field out of the store. Re-entering the page hydrated from the store,
read `undefined`, and rendered **OFF while the backend held ENABLED**.

The second half is what made it unrecoverable: the mutation sends the field only when it
*differs* from the hydrated baseline. With the field missing that baseline was `false`, so after
turning injection ON the operator **could not turn it OFF** -- form said OFF, baseline said OFF,
nothing differed, nothing was sent. Only a hard reload broke the loop.

An operator could enable the disclosure and be told by this screen that it was disabled, on the
one control whose entire job is to say whether exports carry it.

**Grafting, not `fetchMyOrganizations`** -- one, not both, as instructed. It is the same claim
the save path already makes one line later when it advances its own baseline; a re-fetch would
assert no more than that while adding a round trip that can fail on its own and leave the store
stale after a *committed* save. Nothing is grafted when the field is unsupported: writing
DISABLED on a CE mirror would state a choice nobody made.

The graft and the read are extracted to `utils/orgSettingsCommit.ts` **so the round trip can be
tested rather than scanned for**. Every individual piece of this already worked -- the mutation
sent the right value, the server stored it, the local baseline advanced. The failure was in the
shape of what went into the store, visible only on the next hydration, which is exactly what a
unit test of any one function cannot see. The test replays save -> commit -> re-hydrate through
`UPDATE_ORGANIZATION` verbatim. Reverting the fix fails 2 of 5, including the unsendable-DISABLED loop.

## (b) Placement

The switch rendered **between** "Risk-transfer process reference" and "Risk increases over
time" -- inside the three labeling statements, reading as a fourth piece of labeling text and
splitting the three apart. Now its own "Support disclosure export" heading after all three, and
out of the form-item label column that had put it at the far right with 760px between it and
the words it controls.

## (c) Export gate

A refusal knowable before the click belongs on the control, not in a dialog after it. **Only**
the product gate moves: the data-dependent refusals (unresolvable org, failed collect,
unauthored prose slot) keep their modal, because disabling a button for a reason nobody has
checked yet would be a worse lie than the redundant dialog. Tested both ways.

## (d) Withdrawn chip, and the silent Save

"Unknown" would have been accurate about the support state and silent about the fact that a
claim once existed and was retracted -- different facts, and an auditor reads this screen for
both. The EOS suffix is suppressed with the status: printing `EOS 2025-12-31` beside
"Withdrawn" puts the retracted date back on screen as though it stood.

`attestationState` added to the components query -- without it the UI cannot tell a withdrawal
from a component nobody assessed. Placed in CORE, not FULL: **the CE schema already declares it
on `SbomComponent`** (checked, not assumed), so this does not widen the Pro-only surface.

`errors()` has always produced good sentences for every gate; nothing rendered them near the
control they gate. In the run **two fired at once** -- the mandatory re-assert reason, far
enough up the form to be below the fold, and the unacknowledged "does the recorded basis still
hold?" prompt -- so filling only the reason left Save just as dead with nothing saying which
requirement remained. A requirement stated only where the operator is not looking is not stated.

## (f) The statement

The "Software components whose support ends sooner" section named every component and its date
-- on the walkthrough device, all ten. That is a component inventory by another name, in the one
document that must not be one (FDA L1591-1592 puts patients and caregivers in its audience). A
count is no better: "10 of 10" reads worse than "3 of 40" to a reader with no way to compare
them.

**The section stays**, and the file header now records why it is a deliberate deviation from
plan 7f rather than an oversight: a reader told only the device's end-of-support date would
reasonably conclude every part is supported until then -- false whenever a component's window
closes earlier, which is the gap 524B exists to close. 7f's constraint rules out the *detail*,
not the fact.

**Worth recording: revert-probing caught one of my own new tests being theatre.** Restoring the
per-component table failed only 1 of the 3 new assertions -- the section slice was cutting at
the heading's own flattened style marker, so two were asserting against an empty string. With
the slice corrected all three fail on the reverted code.

## Gates

**746 tests across 52 files, all passing** (was 719 -- 27 new). Build clean. Every new guard
revert-probed against the exact defect it claims to catch.

### DIFF
--- a/ui/src/components/OrgSettings.vue
+++ b/ui/src/components/OrgSettings.vue
@@ -994,36 +994,46 @@
                             </div>
                         </n-form-item>
 
-                        <!-- The export toggle sits WITH the FDA text because it is the same
-                             decision: what a generated document or export says about support.
-                             Default OFF (D3), so this is the control a manufacturer preparing
-                             a submission must deliberately turn on. -->
-                        <!-- Hidden entirely on a backend that does not declare the field --
-                             a CE mirror before the deferred sync. Offering a switch whose
-                             write the server would reject is worse than not offering it. -->
-                        <n-form-item v-if="supportInjectionSupported"
-                            label="Carry support attestations in BOM exports">
-                            <div style="display: flex; flex-direction: column; width: 100%;">
-                                <n-switch v-model:value="supportInjectionEnabled"
-                                    :disabled="savingOrgSettings" />
-                                <span class="text-muted" style="margin-top: 4px; max-width: 760px;">
-                                    Off by default. When on, BOM exports carry this
-                                    organization's support attestations &mdash; the artifact
-                                    download, the SPDX-augmented download and the release SBOM
-                                    export. The raw artifact download never carries them.
-                                    <strong>Forged support properties are always removed,
-                                    whatever this is set to</strong>; this controls only
-                                    whether our own attestations are added.
-                                </span>
-                            </div>
-                        </n-form-item>
-
                         <n-form-item label="Risk increases over time (labeling)">
                             <n-input v-model:value="orgSettings.fdaRiskIncreasesNotice"
                                     :disabled="savingOrgSettings" :maxlength="FDA_PROSE_MAX_LENGTH" show-count
                                 type="textarea" :rows="3" style="max-width: 760px;"
                                 placeholder="Cybersecurity risk to users can be expected to increase after end of support." />
                         </n-form-item>
+
+                        <!-- ITS OWN SECTION, AFTER all three labeling slots.
+                             It previously sat between slot 2 and slot 3, which read as a
+                             fourth piece of labeling text and broke the three apart -- an
+                             operator working down the FDA labeling statements met a switch
+                             about export behaviour in the middle of them. The three slots are
+                             prose that ships INSIDE documents; this decides whether a
+                             document carries attestations at all. Related, not the same kind
+                             of thing, and the ordering now says so. Reported by an operator
+                             running the walkthrough, board t20260909-061338-23148. -->
+                        <!-- Hidden entirely on a backend that does not declare the field --
+                             a CE mirror before the deferred sync. Offering a switch whose
+                             write the server would reject is worse than not offering it. -->
+                        <div v-if="supportInjectionSupported" style="margin-top: 18px;">
+                            <h4 style="margin-bottom: 2px;">Support disclosure export</h4>
+                            <!-- The switch is ADJACENT to the words it controls, not in a
+                                 form-item label column that pushed it to the far right with
+                                 760px of whitespace between the two. A control that far from
+                                 its label is one an operator has to aim at. -->
+                            <n-space align="center" :size="10" style="margin: 8px 0 4px;">
+                                <n-switch v-model:value="supportInjectionEnabled"
+                                    :disabled="savingOrgSettings" />
+                                <span>Carry support attestations in BOM exports</span>
+                            </n-space>
+                            <span class="text-muted" style="display: block; max-width: 760px;">
+                                Off by default. When on, BOM exports carry this
+                                organization's support attestations &mdash; the artifact
+                                download, the SPDX-augmented download and the release SBOM
+                                export. The raw artifact download never carries them.
+                                <strong>Forged support properties are always removed,
+                                whatever this is set to</strong>; this controls only
+                                whether our own attestations are added.
+                            </span>
+                        </div>
 
                         <n-divider style="margin: 22px 0 10px;" />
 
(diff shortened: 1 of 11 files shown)


logicflakes:
## Gates complete

Deployed `rearm-pro:t23148-fix2` / `rearm-ui:t23148-fix2` to ns `claude2` (both pods Running,
orphan `kubectl-*` field-managers stripped, cluster lane released), then re-ran the probes.

### Both review layers: 6 + 3 findings, all fixed

**Two were defects this branch introduced, and one of them my own test was defending.**

| Finding | Fix |
|---|---|
| **`deviceSupportRisk` still derived from a WITHDRAWN row** -- a withdrawn component rendered a neutral "Withdrawn" chip beside a red "Outlived by this device -- disclose and address it", while the export omitted it. The exact divergence the commit claimed to close, half-closed | one `withdrawn` flag governs both derived verdicts (`3861af5d`) |
| **The font-coverage whitelist outlived the text it checked.** Removing the per-component list left component names in it, so an unrenderable glyph in a NAME became a hard refusal of the whole patient-facing statement -- and `deviceSupportStatement.spec` asserted exactly that refusal, so the suite defended the bug | whitelist matches what is printed; new case pins that a name cannot block it |
| **The graft was too narrow.** The mutation returns a PARTIAL org and `UPDATE_ORGANIZATION` replaces it, so `type` and `approvalRoles` were also wiped on every settings save -- downgrading the org to DEFAULT for the invite-user form | merges over the stored org |
| **Two error lists.** `errors()` was already rendered in the modal body; my comment said "nothing rendered them" and I had added a second copy | one renderer, in the footer; comment corrected to *moved*, not added |
| Vacuous DISABLED test; CE fixture built through a default parameter that fires on `undefined` | both rebuilt so they can fail |

### Live verification

- **The HIGH fix, with the operator's actual repro:** save ON -> SPA-navigate away and back ->
  **UI ON, backend ENABLED, they agree**; DISABLED sendable afterwards (the unsendable loop is
  gone); `org.type` survived the saves.
  **`toggle-probe.mjs` does not cover this** -- it passed against the broken build -- so a
  separate check was written rather than claiming coverage it does not have.
- **(e) confirmed live:** zero empty `properties` arrays across all 8 served documents.
- Egress table and Layer B unchanged and green on the new image; forged level and forged
  `declarations` absent from all 8 rows.

### Three probe defects found by running them

Committed in `5ad09f14`:

1. **A locator that clicked the wrong switch.** Repairing `toggle-probe` after the toggle moved,
   the obvious `.n-space` filter matched an **outer layout** element holding the page's other
   switches -- so it clicked and **saved `justificationMandatory`** while reporting on it as the
   support toggle, consistently enough to look healthy. Now `.last()`, and the row is asserted
   to hold exactly one switch before anything is clicked. The setting it flipped was restored.
2. **`reset-baseline` claimed a reset it did not perform.** It sent `eol: ''`; empty-string-means-
   clear is the PROSE idiom, and the device window has explicit `clearEos`/`clearEol` flags. The
   mutation succeeded, the script printed OK, and the `2033-09-30` residue stayed. Now sends
   `clearEol` and **reads the release back**.
3. **`spdx-probe` left the org one attestation above baseline on every run** -- it attests a
   shared component and its `finally` archived only the throwaway one it created. Now withdraws
   what it attested.

### Suites

- Backend **2201 tests, 0 failures** (5 pre-existing errors: `t20260907-135917-12775`,
  `t20260908-142935-1289`).
- UI **748 tests / 52 files**, build clean.
- Every new guard revert-probed.

### Sandbox

Left at baseline and **verified on every axis**, not assumed: device window EOS `2031-01-31` /
EOL not declared, release narrative inherited, org coverage **4 of 14**, release coverage
**1 of 10**, export state **DISABLED**.

Untouched per instruction: TLD-ONLY inversion and the product-release "No SBOMs found!"
(`t20260909-105515-31103`). No `dev -> main`, no CE sync.

PR #346, opened 2026-09-10
D7: declare the device support window on the product component (UI)

logicflakes:
UI half of PR 2, for operator decision D7: the section 524B support window is declared on the
**product component** (the device model), optionally overridden per **shipment** (the batch),
and evaluated per **device** (the unit). Backend is rearm-saas #515 (+ #516 for the walkthrough).

**Merge order: rearm-saas #515, then rearm-saas #516, then this, in one sitting.** Dev must
never sit with the backend resolving through the component while the documents still read the
release.

## What moved

| Surface | Before | After |
|---|---|---|
| Product component, Core Settings | -- | **Declares** the window. Renders only when the component has a device class. |
| Release page | Editable `eos`/`eol`, labelled as the device window | Window shown **read-only** with a link to where it was declared; the release's own dates are a **separate, editable** block that says outright it feeds TEA/CLE `END_OF_SUPPORT` / `END_OF_LIFE` events |
| Shipment (Distribution) | -- | Batch **override**, on Hardware/SaMD tabs only |
| Addendum + Device Support Statement | Read the release | Read the component window, and **say where the dates came from** |

`ReleaseData.eos/eol` are untouched and are not renamed or re-documented beyond one line saying
they are no longer the device window. **Nothing is migrated**: copying a release lifecycle date
into a device support window would turn a version's metadata into a labeling claim the
manufacturer never made.

## The CE constraint

CE declares `Component.medicalProfile` **without** `deviceSupportWindow` inside it, so asking
for the subfield makes the **whole document invalid** on CE rather than merely null. Every
document that gained the selection is therefore split CORE/FULL, with a drift spec per
document. `ShippedProduct`/`Device` are SaaS-only and exempt.

`deviceWindowSchemaDrift.spec.ts` also now covers the **shipments** query, which
`validate:graphql` skips because it is built by string interpolation and which nothing else
covered. It uses `it.runIf(proSchema)` rather than an early `return`: a silent green when
rearm-core is absent is how a drift spec stops covering anything without anyone noticing.

## The statement's provenance line

Two statements for the same device model can legitimately carry different dates -- a batch may
override the model default -- and a reader with no way to tell why would reasonably conclude
one of them is wrong. So the statement says, in one plain line under the dates, where they came
from: the product for a model-level window, and site + ship date + lot for a shipment. Never
the word "override", never a uuid: neither is something a patient or a biomed engineer can
check. The shipment form is built and tested but has no entry point yet; that arrives with the
fleet UI, and plan 7h records it so it is not rediscovered as dead code.

## Gates

- `npm run test`: **790 passing, 55 files** (nothing in CI runs this suite; it is part of my
  validation step)
- `npm run build`: clean
- Hot-swapped as `rearm-ui:t7806-d7` and verified live: `d7-window-probe.mjs` **22/22**, driving
  this UI for the release page and downloading the real statement PDF; `toggle-probe.mjs` pass
  with 0 page exceptions.

The live probe matters more than usual here: `ui/` has no `vue-tsc` and `src/utils/` is
unlinted, so a green build and a green suite cannot catch an undefined `<script setup>`
identifier, an invalid GraphQL document or a dead router link. Two of the three Layer 2
blocking findings on this branch were exactly that -- a mutation that would have failed
coercion on **every** save, and a `router-link` to a route `router.ts` does not declare, whose
spec had **pinned the broken form**.

## Out of scope

The fleet UI, shipment-date-anchored windows, `dev -> main`, the CE sync, `5f0daff8`.

### DIFF
--- a/ui/src/components/ComponentView.vue
+++ b/ui/src/components/ComponentView.vue
@@ -526,73 +526,73 @@
                                                  the field -- a CE mirror before the deferred sync.
                                                  Offering an editor whose save the server would
                                                  reject is worse than not offering it. -->
-                                            <!-- The device class itself, editable. Until this
-                                                 existed, deviceClass could ONLY be set at
-                                                 creation (CreateComponent.vue), so the window
-                                                 panel below -- which correctly renders only for
-                                                 a device -- was unreachable for every component
-                                                 that already existed. D7 puts the section 524B
-                                                 commitment on the product component, so the
-                                                 component has to be able to become a device. -->
-                                            <div v-if="deviceWindowSupported" style="margin-top: 22px;">
-                                                <h4>Device class</h4>
-                                                <n-space align="end" style="margin-bottom: 8px;">
-                                                    <n-select v-model:value="deviceClassEdit"
-                                                        :options="DEVICE_CLASS_OPTIONS" style="width: 240px;"
-                                                        :disabled="!isWritable || savingDeviceClass" />
-                                                    <n-button v-if="isWritable" size="small" type="primary"
-                                                        :disabled="deviceClassEdit === deviceClassBaseline"
-                                                        :loading="savingDeviceClass"
-                                                        @click="saveDeviceClass">Save device class</n-button>
-                                                </n-space>
-                                                <div class="text-muted" style="font-size: 12px; max-width: 720px;">
-                                                    Setting this to anything but <strong>NONE</strong> makes the
-                                                    component a device and reveals its support window below.
-                                                    Setting it back to NONE <strong>retracts any declared
-                                                    window</strong>: the window lives in the medical profile
-                                                    that NONE removes.
-                                                </div>
+                                        <!-- The device class itself, editable. Until this
+                                             existed, deviceClass could ONLY be set at
+                                             creation (CreateComponent.vue), so the window
+                                             panel below -- which correctly renders only for
+                                             a device -- was unreachable for every component
+                                             that already existed. D7 puts the section 524B
+                                             commitment on the product component, so the
+                                             component has to be able to become a device. -->
+                                        <div v-if="deviceWindowSupported" style="margin-top: 22px;">
+                                            <h4>Device class</h4>
+                                            <n-space align="end" style="margin-bottom: 8px;">
+                                                <n-select v-model:value="deviceClassEdit"
+                                                    :options="DEVICE_CLASS_OPTIONS" style="width: 240px;"
+                                                    :disabled="!isWritable || savingDeviceClass" />
+                                                <n-button v-if="isWritable" size="small" type="primary"
+                                                    :disabled="deviceClassEdit === deviceClassBaseline"
+                                                    :loading="savingDeviceClass"
+                                                    @click="saveDeviceClass">Save device class</n-button>
+                                            </n-space>
+                                            <div class="text-muted" style="font-size: 12px; max-width: 720px;">
+                                                Setting this to anything but <strong>NONE</strong> makes the
+                                                component a device and reveals its support window below.
+                                                Setting it back to NONE <strong>retracts any declared
+                                                window</strong>: the window lives in the medical profile
+                                                that NONE removes.
                                             </div>
+                                        </div>
 
-                                            <div v-if="deviceWindowSupported && isDeviceComponent"
-                                                style="margin-top: 22px;">
-                                                <h4>Device support window</h4>
-                                                <n-alert type="default" :show-icon="false"
-                                                    style="font-size: 12px; margin: 8px 0 10px; max-width: 720px;">
-                                                    Shown separately and never merged: end of support and
-                                                    end of sale are different facts, and a reader of the
-                                                    Device Support Statement is entitled to both. Blank
-                                                    means <strong>not declared</strong>, which is a fact in
-                                                    its own right &mdash; not an unknown to fill in.
-                                                    <span style="display:block; margin-top:6px;">
-                                                        This is the DEVICE's window, inherited by every
-                                                        release of this product. A release's own end-of-support
-                                                        date is release lifecycle and is set on the release.
-                                                    </span>
-                                                </n-alert>
-                                                <n-space align="end" style="margin-bottom: 8px;">
-                                                    <div>
-                                                        <div class="text-muted" style="font-size: 12px;">End of support (EOS)</div>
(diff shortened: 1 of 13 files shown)


logicflakes:
## Layer 1 (`/review`) -- 8 findings, all confirmed, all fixed

Reviewed both repos' diffs together. Every finding was verified independently against the
source before fixing; none were taken at face value. All 8 were real.

**The headline is worse than any single finding: the entire D7 UI write path was dead on
arrival, and every gate was green.**

| # | Finding | Status |
|---|---|---|
| 1 | `loadInheritedDeviceWindow` read `updatedRelease.component`; `SINGLE_RELEASE_PRODUCT_GQL` -- the query used for PRODUCT releases, the only surface the panel renders on -- does not select it. Read returned early every time; inherited window said "not declared" forever. | fixed, reads `componentDetails.uuid` |
| 2 | `isDeviceComponent` read `componentData.deviceClass`, which neither component query selected. False for every component, so the panel never rendered. | fixed, `deviceClass` added to `COMPONENT_FULL_DATA` (CE declares it, so no CORE/FULL split needed) |
| 3 | `<n-date-picker>` used twice but `NDatePicker` never imported, and naive-ui is not globally registered here. No date inputs even if the panel rendered. | fixed |
| 4 | `collectAddendumData` issued FULL directly, never `loadWithSchemaDriftFallback`. `ADDENDUM_RELEASE_QUERY_CORE` was dead outside its spec -- the CE protection was documented but absent. | fixed |
| 5 | `statementStrings` omitted the SHIPMENT provenance fields, so an unrenderable glyph in a site or lot name bypassed the font-coverage refusal. | fixed |
| 6 | A new shipment always printed "No window is declared for this device model" -- false, and stated exactly when it would talk someone into a needless override. | fixed |
| 7 | `updatedComponent.name` on a `Ref` in `<script setup>` (template unwraps, script does not) sent `name: undefined`. | fixed |
| 8 | Walkthrough told the operator to set Device class on Core Settings -- no such control existed anywhere post-create. | fixed by adding one |

### The root cause Layer 1 did not name

Chasing #2 in the browser turned up the actual reason nothing rendered: both panels were nested
**inside** `<div class="coreSettingsActions" v-if="hasCoreSettingsChanges && isWritable">` -- the
unsaved-changes action bar. The panel appeared only after you had edited some *other* core
setting and vanished the moment you saved. On a freshly opened component -- every reader
following the walkthrough -- it did not exist.

### Why nothing caught it

`ui/` has no `vue-tsc` and `src/utils/` is unlinted, so a clean build and 797 passing tests
cannot see an unregistered component, an unselected GraphQL field, or a query that is never
issued. **And my own live probe was green**, because it wrote the window over GraphQL and
asserted only copy text on the release page -- the copy is static template text and rendered
perfectly while the panel beside it read "not declared".

Layer 1's closing note was right and I am recording it: the source-regex wiring specs are
weaker evidence than the PR body implied, and one of them actively pinned a bug.

### What changed as a result

The probe now drives the **panel**: opens the settings dialog, sets the device class through the
select, fills both date pickers, clicks **Save window**, reads the value back from the server,
and asserts the release page renders the inherited **dates** rather than the sentence.
**30/30 on `rearm-ui:t7806-d7d`.** New specs pin each fix by the property that was violated,
including a div-depth count so the panel cannot drift back inside the action bar.

Suite 797 passing, build clean.

logicflakes:
## Layer 2 (skeptical-senior-ReARM-engineer conventions pass) -- 11 findings, all addressed

Ran against both repos' diffs together, per the persona brief in
`rearm-core/ai-agents/review-layer2-persona.md`.

**Clean, stated plainly:** zero non-ASCII on any added line in either repo (CI exit-123 risk:
none) · no new dependencies, no `package.json`/`pom.xml` touched · eslint 12 errors on this
branch and the identical 12 on `origin/fda-readiness-1-dev`, **zero introduced** (verified by
stashing) · drift-spec shape matches the 8 sibling `*SchemaDrift.spec.ts` files · `deviceClass`
string literals match `CreateComponent.vue:84`, `DeviceView.vue:113`, `DistributionOfOrg.vue:368`
— no shared constant exists to use.

### The three it flagged as pre-merge

**Restated seams instead of using the repo's own.** `componentDeviceWindow.ts` declared an
inline `{ query: (opts: any) => Promise<any> }` where `graphqlDriftFallback.ts:117` exports
`DriftFallbackClient`, and took the drift predicate as a parameter where every neighbour imports
`isSchemaDriftError` directly. Both call sites passed the same function — the seam bought
nothing but a **fake predicate in the spec**, which would have kept passing if the real
classifier stopped recognising a validation error. Now imported; the spec throws the
graphql-java-shaped error CE actually returns.

**Two error extractors 46 lines apart in one feature.** `saveDeviceWindow` used
`parseGraphQLError(e.message)`, which strips three BOM prefixes and never unwraps
`graphQLErrors[]`. Under Apollo's `CombinedGraphQLErrors` the server's own text — "End of
support cannot be after end of life", the device-class refusal **the walkthrough quotes at step
1.2** — never reached the panel. `saveDeviceClass` forty lines up already used the right one.

**The shipment path skipped the extract-to-utils step its sibling took**, then tested the inline
copy by regexing the `.vue`. Two copies of what "retract" means on a 524B commitment. Extracted
to `applyShipmentWindowToInput` / `effectiveWindowLabel` with 10 behavioural assertions. The old
spec pinned the literal ternary `w.source === 'SHIPMENT' ? ... : ...` — renaming a local would
have failed it while a genuinely wrong string passed.

### Also fixed

- A 19-line JSDoc whose load-bearing warning is **"WIDENING IT ARMS THE GATE"** had been
  detached from the `deviceWindow` it warns about when I inserted the inherited-window block
  above it. Reattached. Same shape on `deviceWindowMutationInput` (two stacked blocks, only the
  second attaching) — merged.
- A spec asserting the **whole document** prints no uuid, green only because the fixture used
  `'r-1'`. `flat()` walks `content`, not the footer — so it was never testing the footer, which
  is exactly where the uuid deliberately goes. Both claims now stated separately with a real
  uuid, "no uuid" scoped to the provenance line.
- Dead second assignment to `editingShipment`; moved panels re-indented to their siblings' depth
  (all three now at 40, matching `coreSettingsActions`).

808 tests passing · build clean · eslint unchanged from base · re-probed on `rearm-ui:t7806-d7e`:
**30/30**, toggle 0 page exceptions, baseline read back on all five values.

PR #349, opened 2026-09-11
ui: read-only "Devices at support risk" panel on the Distribution page (D7 fleet view)

logicflakes:
## What

A read-only **Devices at support risk** panel at the top of `DistributionOfOrg`, consuming `devicesAtSupportRisk`: unit by identifier, client, site, release judged (+ `expected`/`reported` tag), device window with provenance, verdict tag (red verdict / green OK / amber "Not assessed" -- never rendered as OK), earliest component EOS, components driving risk. Follows the client / site selected below it; remote paging on the server total; row click opens the device twin.

## Schema drift (CORE/FULL)

`ui/src/utils/fleetSupportRisk.ts` splits the document: CORE = ids + verdict, FULL = + evidence. Because CE does not declare the query at all today, the loader distinguishes **hide** (query absent: CORE rejected too -> `supported: false`, panel not rendered) from **degrade** (enrichment rejected -> CORE served, evidence columns blank + amber notice). Non-drift errors (403, 5xx) are surfaced via `notify`, not swallowed. `fleetSupportRiskSchemaDrift.spec.ts` validates both documents against the CE schema (pins that the QUERY is what CE lacks) and, when `rearm-core` is checked out alongside, against the Pro schema (`it.runIf`).

## Verification

- `npm run test`: **58 files / 829 tests** pass (11 drift/loader/helper + 10 source-wiring specs added); `npm run build` exit 0; `validate:graphql` 0 Pro failures; no invisible unicode.
- **UI-driven probe** `fleet-risk-probe.mjs` (rearm-core PR, `ai-agents/probes/fda-readiness-1/`) ran this branch's `ui/dist` over the live sandbox backend (`rearm-backend:fda-readiness-1-dev.24`): **16/16** -- panel/headline/row values compared to the server's own `devicesAtSupportRisk` answer, client and site rows *clicked*, empty client reads "No in-field units to evaluate", row click lands on `/device/show/<uuid>`. Sandbox restored (`reset-baseline.mjs` all OK).
- Walkthrough step 12 written from the probe run (same rearm-core PR).

## Not in this PR

Backend follow-up worth considering: the service sorts by device uuid, so on a large fleet an at-risk unit can sit on page 3 -- sorting at-risk first server-side would make the headline's "this page" caveat unnecessary.

Do not merge yet -- two-layer review gate outcomes will be noted below.

### DIFF
--- a/ui/src/components/DistributionOfOrg.vue
+++ b/ui/src/components/DistributionOfOrg.vue
@@ -8,6 +8,22 @@
             </n-space>
             <n-data-table :columns="fleetDriftColumns" :data="fleetDrift" size="small" :row-props="driftRowProps" :pagination="fleetDrift.length > 5 ? { pageSize: 5 } : false" />
         </div>
+        <!-- D7 fleet question, read-only: which in-field units outlive their software.
+             Hidden entirely on a backend whose schema lacks the query (CE before the sync);
+             follows the client / site selected below. -->
+        <div v-if="fleetRisk.supported" class="fleetRiskPanel" data-testid="fleet-risk-panel">
+            <n-space align="center" size="small" style="margin-bottom: 6px;">
+                <n-tag :type="fleetRiskPageSummary.atRisk > 0 ? 'error' : 'default'" size="small">SUPPORT RISK</n-tag>
+                <strong data-testid="fleet-risk-headline">{{ fleetRiskHeadlineText }}</strong>
+                <span class="subtle">{{ fleetRiskScopeLabel }}</span>
+            </n-space>
+            <n-alert v-if="fleetRisk.degraded" type="warning" :show-icon="false" size="small" style="margin-bottom: 6px;" data-testid="fleet-risk-degraded-alert">
+                This backend serves the verdict per unit but not the evidence behind it (release judged, window in force, component end-of-support). Those columns are blank, not empty.
+            </n-alert>
+            <n-data-table remote :columns="fleetRiskColumns" :data="fleetRisk.rows" :loading="fleetRiskLoading" size="small"
+                :row-props="fleetRiskRowProps" :row-key="(r: any) => r.device"
+                :pagination="fleetRiskPagination" @update:page="loadFleetRisk" />
+        </div>
     <div class="distWrapper">
         <!-- Clients column -->
         <div class="distColumn">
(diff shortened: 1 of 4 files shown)


logicflakes:

Layer 1 (`/review`, correctness) and Layer 2 (ReARM-conventions pass, persona `ai-agents/review-layer2-persona.md`) both ran on 0541c9f2; consolidated findings and what happened to each, in 3c76cbe8:

| # | Finding | Layer | Outcome |
|---|---------|-------|---------|
| 1 | A non-drift error (403, timeout) set `FLEET_RISK_UNSUPPORTED`, hiding the panel for good; the watch never retried | L1 | **Fixed** - `fleetRiskError` + red alert with Retry over the last rows; only `supported: false` hides |
| 2 | No request sequencing: a slow org-wide page could overwrite a newer client scope | L1 | **Fixed** - monotonic ticket in `loadFleetRisk`, only the newest writes |
| 3 | `(from the product component)` on every row - backend `windowDto` serialises `source` as null, `effectiveWindowLabel` defaulted the provenance | L1+L2 | **Fixed** - `source` dropped from the selection (pinned in spec), `fleetWindowLabel` carries dates only |
| 4 | Tooltip reused `DEVICE_RISK_DETAIL` ("this component...") on a unit row | L2 | **Fixed** - `FLEET_RISK_DETAIL: Record<FlaggedDeviceRisk, string>` |
| 5 | `=== 'REPORTED'` magic strings in the render | L2 | **Fixed** - `RELEASE_SOURCE_TAG` / `releaseSourceTag()` |
| 6 | Unrecognised verdict rendered as "Not assessed" | L2 | **Fixed** - `UNRECOGNISED_TAG` + `console.error`, counted separately in the headline (same rule as `supportTag`) |
| 7 | "No in-field units to evaluate" flashed before the first load | L1 | **Fixed** - `fleetRiskLoaded` gates the panel |
| 8 | `fleetRiskFullRejected` redundant with `degraded` | L2 | **Fixed** - `skipFull: fleetRisk.value.degraded` |
| 9 | `resolveSiteInfos` guard re-queried on an empty org and never invalidated | L1 | **Fixed** - loaded flag, invalidated on save/archive site |
| 10 | Headline "evaluated" overstates: the server evaluates only the returned page | L1+L2 | **Fixed** - "in scope" |
| 11 | `fleetRiskColumns` needlessly computed; repeated em-dash spans; spec brittleness (literal arrow pin, notify title) | L2 | **Fixed** - plain const, `blankCell`, specs pin `name: 'DeviceView'` / `deviceuuid: r.device` separately, title `'Failed'` |

Validation after fixes: `npm run test` 835/835 (58 files; the two new specs are 27 of them), `npm run build` green, no invisible unicode. Live UI-driven probe `fleet-risk-probe.mjs` (rearm-saas#522) 17/17 against this branch's `dist` served over the deployed sandbox backend, then `reset-baseline` all OK.

**Follow-ups, not this PR** (backend, rearm-core):
- `devicesAtSupportRisk` rows should carry identifiers + `siteName`/`clientName` server-side like `FleetDriftEntry` does; the UI currently does one `devicesOfSite` per distinct site on the page to label units.
- Sort at-risk first and/or return `atRiskTotal`, so page 1 of a large fleet answers the question instead of a page count.
- `attestedEosFor` does two repo reads per row; a per-release memo within one request would collapse that.
- `isSchemaDriftError` treats any 400 as drift (shared primitive, pre-existing).

PR #350, opened 2026-09-11
ui: fleet-risk rows carry their own labels and the fleet-wide at-risk count

logicflakes:
Consumes the row fields relizaio/rearm-saas#524 added, on the Distribution page's "Devices at support risk" panel (D7).

## What changes

| | before | after |
|---|---|---|
| Unit column | `devicesOfSite` per DISTINCT site on the page, then look the unit up | `identifiers` on the row |
| Client / Site | org-wide `sitesOfOrg` + the loaded client list | `siteName` / `clientName` on the row |
| Headline | `M in-field units in scope; this page: A at risk, B not assessed, C OK` | `K at risk of M in-field units in scope`, K = `atRiskTotal` over the whole filtered fleet |
| Order | server sorted by unit | server sorts at-risk-first across the fleet; rendered as served, never re-sorted |

Rendering the org-wide panel is now **one** roster read instead of one plus N site reads.

## Drift pattern kept

The new fields are FULL-only, alongside the evidence columns, read behind the same `'field' in r` presence guard. A CORE-served page still renders the roster (units named by short id) and falls back to the page-count headline. `atRiskTotal` is `number | null`, never coerced to 0: an unknown fleet count rendered as "0 at risk" would be the panel asserting a disclosure fact it was not told. The degraded alert now names the labels and the count among what that backend does not serve.

Null `siteName` / `clientName` on a FULL page means no placement **or an archived site** (backend behaviour); the row still renders, with an em dash.

## Validation

- `npm run test`: **58 files, 840 tests, all passing** (no CI runs this suite). New coverage: FULL carries `atRiskTotal` + labels; `atRiskTotal` stays null when the server omits it, nulls it, or the page is degraded; the "K at risk of M" headline including the fleet count overriding what the page shows; the page-count fallback; and wiring pins for the row-sourced labels, the absence of `devicesOfSite` / `sitesOfOrg`, and the absence of any client-side sort.
- `npm run build`: clean.
- Browser-render check, branch build served over the sandbox (`DIST=ui/dist`, backend `fda-readiness-1-dev.25`, i.e. post-#524): `fleet-risk-probe.mjs` **22/22**, including three new assertions that fail against the deployed UI by design. Probe change is relizaio/rearm-saas (branch `probe/fleet-risk-ui-labels-headline`).

Do not merge yet: two-layer review gate to follow.

### DIFF
--- a/ui/src/components/DistributionOfOrg.vue
+++ b/ui/src/components/DistributionOfOrg.vue
@@ -25,7 +25,7 @@
                 </n-space>
             </n-alert>
             <n-alert v-if="fleetRisk.degraded" type="warning" :show-icon="false" size="small" style="margin-bottom: 6px;" data-testid="fleet-risk-degraded-alert">
-                This backend serves the verdict per unit but not the evidence behind it (release judged, window in force, component end-of-support) nor the unit labels (identifiers, site and client names) or the fleet-wide at-risk count. Those columns are blank, not empty; units are named by id and the headline counts this page only.
+                This backend serves the verdict per unit but not the evidence behind it (release judged, window in force, component end-of-support) nor the unit labels (identifiers, site and client names) or the fleet-wide at-risk count. Those columns are blank, not empty; units are named by id, and the headline counts the rows in hand rather than the fleet.
             </n-alert>
             <!-- Served at-risk-first across the whole fleet; rendered in that order, never re-sorted. -->
             <n-data-table remote :columns="fleetRiskColumns" :data="fleetRisk.rows" :loading="fleetRiskLoading" size="small"
@@ -874,6 +874,9 @@ const fleetRiskPagination = computed(() => ({
 // FULL page with a null name is a unit at an archived site, shown by id so it is not lost.
 const fleetRiskUnitLabel = (r: FleetRiskRow) => summarizeIds('identifiers' in r ? (r.identifiers || []) : []) || shortUuid(r.device)
 const fleetRiskSiteName = (r: FleetRiskRow) => ('siteName' in r && r.siteName) || (r.site ? shortUuid(r.site) : '')
+// No id fallback, unlike the site: the row carries no client uuid, so a unit whose CLIENT was
+// archived reads the same as a site with no client at all. The panel will not invent a
+// distinction it was not given; closing it needs `client: ID` on the row, backend-side.
 const fleetRiskClientName = (r: FleetRiskRow) => ('clientName' in r && r.clientName) || ''
 // Every load takes a ticket; only the newest ticket may write. A client switch while a
 // slow org-wide page is in flight would otherwise land the org-wide rows under the client's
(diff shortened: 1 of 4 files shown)


logicflakes:

**Layer 1 (`/review`)** -- 3 findings, all addressed in c00b14b7:

1. *Widening the single FULL document means a Pro backend that has the query but not the #524 fields degrades all the way to CORE, blanking evidence it could serve, and `skipFull` makes that sticky for the session.* Accepted as a deliberate trade, now written down in `fleetSupportRisk.ts` rather than left for the next reader: the alternative is a third drift tier that costs an extra round trip on every fallback, forever, to buy a partial answer during the one deploy where the UI is ahead of the backend. A reload after the backend rolls restores FULL.
2. *`fleetRiskClientName` has no id fallback, so a unit whose CLIENT was archived reads the same as a site with no client -- the old `sitesOfOrg` read supplied `shortUuid(client)`.* Real, and not closable UI-side: the row carries no client uuid. The panel will not invent a distinction it was not given; the comment now names the backend fix (`client: ID` on `DeviceSupportRiskRow`). Logged as a follow-up on the board task.
3. *The degraded alert claimed the headline "counts this page only", but the CORE fallback reports fleet-wide counts when the fleet fits on one page.* Fixed -- it now says it counts the rows in hand rather than the fleet.

**Layer 2 (ReARM conventions persona)** -- 4 findings, all fixed in 127112956:

1. *The FULL headline dropped `unknown` entirely, so `0 at risk of 57` over a fleet where 50 units have no declared window read as "clean" -- and the degraded path disclosed MORE than the healthy one.* The file's own doctrine (`fleetRiskTag`) says a unit nobody has looked at is not a safe unit. The headline now names them as the page fact they are: `0 at risk of 57 in-field units in scope; 20 not assessed on this page`. The probe lost its only assertion on non-flagged wording when the old check went; a replacement went in (rearm-saas#525).
2. *`idType: string` where the wire carries the `IdentifierType` enum*, in a file that mirrors its other two enums as unions precisely so a `.vue` render cannot compare against a display spelling. Mirrored, with a spec pinning the union against BOTH schemas so a new identifier type cannot land unmirrored.
3. *Two wiring pins tighter than the file's own calibration* (a whole `<n-tag>` element including attribute order; the label expression character-for-character). Loosened to the binding and the call shape.
4. *The no-re-sort assertions could pass vacuously* -- a reworded section marker slices to `''` and both `not.toMatch` calls pass while covering nothing. Positive assertion added on the same slice, matching the idiom everywhere else in that file.

Also caught by Layer 2 and worth recording: nothing was left half-removed (`sitesOfOrg` is gone entirely, `devicesOfSite` survives only in the site-selection path), the diff is plain ASCII, and no new dependencies.

## Validation after the fixes

- `npm run test`: **58 files, 842 tests passing** (nothing in CI runs this suite; it is part of my validation step).
- `npm run build`: clean.
- `fleet-risk-probe.mjs` against sandbox backend `fda-readiness-1-dev.25` with this branch's build served over the origin (`DIST=ui/dist`): **23/23**.

PR #351, opened 2026-09-11
ui: generate the Device Support Statement from a shipment (plan 7h)

logicflakes:
Wires the entry point plan 7h deferred: **a Device Support Statement generated from a SHIPMENT**, as an action on the shipment row of the Distribution page. Design note on board task `t20260907-044617-15432`; this implements it with the operator's three conditions.

The SHIPMENT provenance shape (`AddendumData.deviceWindowSource`, site / ship date / batch identifier) and its renderer have existed with tests since the statement landed -- reachable only from tests, because nothing produced it. Nothing about the document changes here; what changes is which inputs are collected.

## The three conditions

**1. Provenance and the effective window travel together.** `ShipmentStatementContext` carries `siteName` / `shipDate` / `batchIdentifier` **and** `eos` / `eol` in one object, so a caller cannot supply one without the other. `collectAddendumData` takes the dates from it when it is present, and **refuses** a delivery with no window in force rather than falling through to the model default. That fallback is the whole failure mode: a line naming a batch printed over the dates that batch overrode -- a statement contradicting itself about which units it describes.

Unit test, as asked, asserting on the **rendered document** rather than the collected facts (only the document can show the two disagreeing): the fixture's model window is `2031-01-31` / `2034-12-31`, the delivery's is `2029-03-31` / `2030-09-30`, and the PDF definition carries the delivery's and not the model's.

**2. The row action is gated on an effective window.** No window in force, no action: the icon is inert with the reason on it, rather than a button that answers the click with a dialog -- the pattern the release view already had to walk back (board `t20260909-061338-23148` step 6d).

Gated on the **window in force, not on the shipment's classification**: a plain-software delivery of a device model reached units that run under a window too. So both shipment tables offer it, from one shared column definition. That also removes a live trap -- the software table picked the edit icon up as `shipmentColumns[length - 1]`, which would have silently handed it whichever column was appended last.

**3. The statement names the batch identifiers and the source.** `batchIdentifier` is the shipment's identifiers joined the way this page shows them everywhere else (`summarizeIds`), so the provenance line reads `These dates apply to the units delivered to St Elsewhere ICU on 2026-04-02 (batch LOT:77, SERIAL:A9), and may differ from other deliveries of the same device.` Still never the word "override" and never a uuid.

## Also in here

`collect / refuse / render / download` moved whole into `utils/deviceSupportStatementExport.ts`. Two callers now ask for this document, and a second copy of that sequence is a second chance to skip the block check -- which is what keeps a patient-facing PDF with an empty section from existing. The refusals are **returned, not shown**: the release view keeps its modal and the distribution page notifies, and both still distinguish FAILED ("something went wrong") from BLOCKED ("this document must not exist yet"). The wiring pins that used to read `ReleaseView.vue` moved to the helper rather than being deleted.

## Validation

- `npm run test`: **58 files, 853 tests passing** (nothing in CI runs this suite). New: the four shipment cases above, plus wiring pins for the shared helper, the shipment context, the gate, and both tables reading one column.
- `npm run build`: clean.
- Live, against sandbox backend `fda-readiness-1-dev.25` with this branch's build served over the origin: `d7-window-probe.mjs` **38/38**, including a new step 6.7b that downloads the shipment-generated PDF and reads it with `pdftotext` -- batch dates present, model dates absent, the delivery named, no "override", no uuid. Probe change is relizaio/rearm-saas#<probe PR>.
- Sandbox restored and read back: shipment `deviceSupportWindow` and `effectiveDeviceSupportWindow` both `null`, as before the run.

One honest gap: the live probe exercises the **omit** branch of the batch clause, not the naming one -- the sandbox's only shipment is plain software, and the server clears batch identifiers on those. The naming branch is unit-tested.

Not in this PR, still open on the board: `client: ID` on `DeviceSupportRiskRow`, so a unit whose client is archived stops reading as a unit with no client.

Do not merge yet: two-layer review gate to follow.

### DIFF
--- a/ui/src/components/DistributionOfOrg.vue
+++ b/ui/src/components/DistributionOfOrg.vue
@@ -249,7 +249,7 @@ import { useStore } from 'vuex'
 import { useRoute, useRouter } from 'vue-router'
 import { NTabPane, NTabs, NDataTable, NModal, NForm, NFormItem, NInput, NInputNumber, NSelect, NButton, NIcon, NTag, NSpace, NDivider, NDynamicInput, NTooltip, NPopconfirm, NDatePicker, useNotification, NotificationType, NCheckbox, NAlert } from 'naive-ui'
 import { CirclePlus, InfoCircle, Edit as EditIcon, Archive as ArchiveIcon,
-    FileCertificate as StatementIcon } from '@vicons/tabler'
+    FileCertificate as StatementIcon, Loader as PendingIcon } from '@vicons/tabler'
 import gql from 'graphql-tag'
 import graphqlClient from '../utils/graphql'
 import commonFunctions from '@/utils/commonFunctions'
@@ -261,6 +261,7 @@ import { loadDevicesAtSupportRisk, fleetRiskTag, fleetRiskHeadline, summarizeFle
     type FleetRiskResult, type FleetRiskRow } from '@/utils/fleetSupportRisk'
 import { isDeviceRiskFlagged } from '@/utils/supportStatusTag'
 import { generateDeviceSupportStatement } from '@/utils/deviceSupportStatementExport'
+import { NO_WINDOW_IN_FORCE } from '@/utils/addendumData'
 
 const route = useRoute()
 const router = useRouter()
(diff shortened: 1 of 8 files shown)


logicflakes:

Both layers ran over both diffs (this and relizaio/rearm-saas#526). Everything below is fixed in `76076312` / `69ce4be6` unless it says otherwise.

### The one that mattered

**The provenance was not following the dates.** A delivery that *inherited* the model window still got the SHIPMENT line -- "these dates apply to the units delivered to X ... and may differ from other deliveries" -- over dates that are the model's statement about every unit. Two patient-facing documents for the same device would have contradicted each other about scope, and it was the **common** case, not an edge one: the PR's own live probe only escaped it because it wrote an override first.

`ShipmentStatementContext` now carries `windowSource` from the server's resolver (`effectiveDeviceSupportWindow.source`, already selected on the row), and the provenance branch keys on it. Both layers caught it independently; Layer 2 also caught that the wiring pin enumerating the context fields did not list `windowSource`, so the one spec family whose stated job is "a `<script setup>` identifier nothing type-checks" had the same hole. Pin and test added.

### Layer 1 (`/review`), remaining 5

1. **File name collision across deliveries** -- two batches of release 1.4.2 both downloaded as `device-support-statement-1.4.2.pdf` with different dates inside, in a folder of regulatory documents. The name now carries the delivery when the dates are a delivery's, derived from the data so it cannot disagree with the provenance line inside.
2. **The no-window refusal was classified FAILED** -- a red "something went wrong" for a normal data state, from the one code path built to distinguish that from BLOCKED. Now BLOCKED, and answered before the component walk rather than after it (Layer 2 raised the same ordering point).
3. **"Throws nothing" was false** -- `renderDeviceSupportStatementBlob` throws and was awaited outside any try. Wrapped.
4. **`statementExportPending` was never surfaced** -- a multi-query walk looked like a dead control. The row that was asked shows the work (per-row, not a bare boolean that would spin every row).
5. **The column appeared for orgs that declare no windows at all** -- a permanently inert icon whose tooltip gives device-model instructions. It now appears per table only once some delivery there has a window; the inert icon on its neighbours is then the signal it was written to be.

### Layer 2 (ReARM conventions), remaining

- **The tooltip and the refusal were two sentences that must agree, in two files** -- one exported constant now.
- **The site name came from the page-level `selectedSite`** while every other field came from the row. The context exists so one delivery's provenance cannot be paired with another's window; that was true only by accident of `loadShipments` being site-scoped. Read from the row.
- **The last positional column borrow** (`shipmentColumns.value[1]`) is gone -- the release column is a named const, same as the two this PR already extracted. The spec now forbids the pattern rather than one instance of it.
- **Three pins keyed on whitespace or on an untouched query string** (exact spacing inside `SHIP_FIELDS`, four-space indentation). Loosened to what they guard.

**Reviewed and declined, with the reason now in the code:** `batchIdentifier` entering `addendumData.ts` as a pre-joined string, against that module's "collect facts, render nothing" header. The field it feeds, `DeviceWindowProvenance.batchIdentifier`, is already a single reader-facing string from 7f; re-deriving it downstream would make the document's own provenance line the second source of truth for it. The header forbids formatting facts *collected from the server* -- this is a caller's statement about its own row.

Also **checked and cleared** by Layer 2, worth recording: the font-coverage whitelist already covers the SHIPMENT branch's three user-entered strings, so the joined `TYPE:value` string is inside the refusal; no refusal is bypassed on the shipment path; `isWritable` correctly does not gate a read/export action; nothing is left dead in `ReleaseView.vue` and its FAILED/BLOCKED split is preserved verbatim.

### Validation after the fixes

- `npm run test`: **58 files, 858 tests passing**.
- `npm run build`: clean.
- `d7-window-probe.mjs` against sandbox backend `fda-readiness-1-dev.25` with this branch's build: **38 pass, 1 skip** -- the skip is honest, see the probe PR.
- `fleet-risk-probe.mjs` re-run after the shared-module refactor touched it: **23/23**.
- Sandbox read back to baseline: shipment window null, `devices` / `identifiers` / `choiceResolutions` intact.

PR #366, opened 2026-09-18
fix(helm): require the oci-registry secret instead of silently starting without it

logicflakes:
Re-cut of #325 onto `main`, carrying **only** the helm fix. Original commit and
authorship preserved (`32674588` -> `e3690a0f`, ReARM Agent (claude2)); #325 is
being closed in favour of this.

**Why re-cut:** #325 said in its own description that the fix is not FDA-related
and targets `main` directly -- but its branch was cut from the FDA work, so the
PR carried **15 commits / 89 files / 13.5k insertions** of FDA-Readiness
attestation UI and a CE backend sync, and conflicted on `ReleaseView.vue`. Those
FDA commits already live on the `2026-09-7*` branches; they do not need this PR.
This one is 2 files, 19 lines.

## The problem

`oci-artifact-service` exits fatally when `REGISTRY_USERNAME`, `REGISTRY_TOKEN`
or `REGISTRY_HOST` is empty. **It has no anonymous mode**, so there is no
configuration in which an absent `oci-registry-secrets` is survivable.

The chart marked both `secretKeyRef`s `optional: true`, so a missing Secret did
not stop the pod starting: it started with empty credentials and died, and kept
dying. Seen in a live deployment at **1560 restarts**, surfacing only as
`CrashLoopBackOff`, with the real cause (`create_secret_in_chart: false` and no
externally-managed Secret) visible nowhere but the container log.

It took BOM ingest down with it -- rebom validates and dedups the BOM, then
cannot push it and `addBom` fails -- and what reached the user on upload was:

```
BOM processing failed: Rebom error: fetch failed
```

Four services away from the fault, naming none of it.

## The change

Drop `optional: true`, so Kubernetes reports the fault on the pod in the words of
the thing that is actually wrong:

```
CreateContainerConfigError: secret "oci-registry-secrets" not found
```

`values.yaml` states the contract where someone choosing it will read it: `false`
means you supply the Secret, there is no anonymous mode, and `bundledRegistry` is
the option if you have no external registry.

## Validation (re-run on this branch, not inherited)

- `helm lint` clean.
- **External path** (`create_secret_in_chart=false`): both refs render with no
  `optional`, which is the change.
- **Default** (`create_secret_in_chart=true`): chart still emits
  `oci-registry-secrets` and wires it.
- **Bundled path** (`bundledRegistry.enabled=true`): unaffected -- still resolves
  to `<release>-zot:5000` with `<release>-zot-credentials`.


logicflakes:
## Reviewed, findings fixed in `6034dce9`

Four findings, none in the rendered output -- all in the prose, which on a
19-line chart change is where the risk actually lives.

**1. The helper's own header still argued the opposite.** Directly above the refs
this PR makes required, the comment explained that the external-registry secret
is optional "as it has always been", because a deployment managing it outside the
chart should not be blocked from starting. That is the exact reasoning someone
would use to re-add `optional: true` and restore the silent crashloop. Rewritten
to say it is required and why the old rationale does not hold.

**2. It does not exit at startup -- I had that wrong, including in this PR's
description.** The `Fatal` is inside `healthCheck`, wired at
`r.GET("/health", healthCheck)` (`main.go:37`, check at `:315-323`). So the
process **starts**, the liveness probe hits `/health`, and it dies there. Same
`CrashLoopBackOff` outcome, but an operator told "the pod will not start" who
then watches it start has been sent looking in the wrong place. Corrected in the
template comment and `values.yaml`, with the mechanism named.

**3. `values.yaml` did not say that `create_secret_in_chart: true` is ignored
when `bundledRegistry.enabled` is true** -- that path uses the zot credentials
and the chart creates no `oci-registry-secrets` at all (the Secret template ANDs
`not bundledRegistry.enabled`). Choosing between these two settings is exactly
when someone reads that block.

**4. Not fixed here, and it needs a decision: the Pro chart is not mirrored.**
`rearm-core/deploy/rearm-helm/templates/_helpers.tpl:136-144` still has
`optional: true`, so the two charts are no longer diffable the way this helper's
comment claims. The mirror is not a copy-paste, because in Pro the helper has a
**second consumer**: `oci-artifacts-backup-cronjob.yaml:64`. Making the refs
required there means backup Job pods go to `CreateContainerConfigError` when the
Secret is absent instead of running with empty credentials -- arguably the same
improvement, since the backup needs those credentials too, but it changes Job
behaviour and belongs in its own PR against `rearm-saas`.

Re-validated after the edits: `helm lint` clean, external render still emits no
`optional`, default and bundled paths unchanged.

logicflakes:
Live check on k3s sandbox (ns claude2, which has no oci-registry-secrets). I applied the rendered oci-artifact Deployment from both charts with `create_secret_in_chart=false`:
- **Before (main):** `Error` -> `CrashLoopBackOff` within 14s. The log shows `FATAL build/main.go:309 REGISTRY_USERNAME or REGISTRY_TOKEN or REGISTRY_HOST not set`.
- **After (this PR):** `CreateContainerConfigError`, and the pod events show `Error: secret "oci-registry-secrets" not found`, with 0 restarts.

The probe deployments have been deleted. The CE chart (rearm#366) renders the same `rearm.ociRegistryEnv` block, so this result covers it too.

PR #372, opened 2026-09-20
feat(rebom): processed BOMs carry their own serial number

taleodor-claude:
## Summary

`serialNumber` + `version` is the identity of **one** CycloneDX document. rebom writes several documents per upload — the producer's bytes, the augmented copy, and one more for every enrichment run — and all of them claimed the producer's identity. That made a BOM-Link `urn:cdx:<serial>/<version>` ambiguous, left a consumer holding a cached augmented copy unable to tell it from a later enrichment of the same row, and made a verifier comparing the augmented download against the producer's signature fail on a document the producer never claimed.

Every document rebom pushes now gets its own serial, minted immediately before the push, and carries a `type: bom` external reference back to the producer's identity so the provenance survives the split.

Complements rearm-saas#546, which fixes the raw side (uploaded bytes served under an `AS_UPLOADED` digest). Independent; land in either order.

## Changes

**What does not move.** `meta.serialNumber` stays the producer's. It is what `bomBySerialNumber`, `allBomsBySerialNumber`, `findBomBySerialNumberAndVersion`, duplicate detection and the version-conflict check all key on, and what rearm-core stores as `InternalBom.id`. The `version` field inside the processed document stays the producer's too: `ArtifactService.getArtifactBomLatestVersion` reads the artifact's latest version back out of that body, so the processed copy cannot carry a counter of its own. A fresh serial is spec-correct without one — version is only meaningful within a serial.

- `bomProcessingService`: new `mintProcessedSerialNumber` / `bomLink` / `producerBomLink`. `augmentBomForStorage` mints for the document it returns; augmentation done to *serve* a request (`bomSearchService`, `bomDiffService`) still calls `augmentBomWithComponentContext` directly and mints nothing.
- Both enrichment push sites mint a fresh serial per run — minting once per row would reintroduce the ambiguity on the second enrichment — and refresh the producer reference rather than appending a second one.
- `bomAddService`: mints and records `meta.processedSerialNumber` for every **processable** upload — not for every *augmented* one. Augmentation is not what makes the stored copy differ from the producer's; `processBomObj` has already sanitized, deduplicated and repaired dependencies by then. Gating on augmentation would mean flipping `AUGMENT_ON_STORAGE` silently republished a deduplicated document under the producer's serial. The non-processable (CDX 2.0+) path is left alone: that copy really is the uploaded bytes, verbatim, so it keeps the uploaded identity. The SPDX-converted CycloneDX is also a distinct instance and gets its own serial, with **no** source reference: BOM-Link has no form for an SPDX document, and no link beats an invented one.
- `updateEnrichmentStatusWithBom` carries the serial alongside the tag in the same statement, so the pointer and the identity it describes cannot disagree; each `EnrichmentRun` entry records the serial of the document that run pushed (entry 0 is the upload).
- `schema.graphql.ts`: `BomMeta.processedSerialNumber` and `EnrichmentRun.serialNumber`.
- Merge output is out of scope — merged BOMs already receive their own serial.

**Not included, deliberately:** the optional read-only `processedSerialNumber` on rearm-core's artifact GraphQL type. Nothing in rearm-core needs it to function; happy to add it in a follow-up if the UI wants to show it.

## Testing

- `test/unit/processedSerialNumber.test.ts` (17 new tests): a fresh serial per mint; `version` untouched; the input document and the `externalReferences` array it may share with the raw copy are not mutated; the producer link is built from the document's own identity on the first pass and refreshed — never duplicated — on later passes; `null` source link writes no reference; an accepted upload pushes raw and processed under different identities with the producer link present and `meta.serialNumber` unchanged; the pushed document still passes `validateBom` against the CycloneDX schema; the pointer write records the serial in `meta` and on the run entry.
- The mint's independence from augmentation is pinned: an upload with `augmentBomForStorage` stubbed to a passthrough still pushes a minted, back-linked document and records `processedSerialNumber`.
- The producer reference is shown to stay schema-valid at spec **1.3**, which predates BOM-Link (`bom` has been an external-reference type since 1.2; `url` is an unconstrained string). Asserted on a bare document rather than an augmented one — see the reviewer note below.
- Legacy rows: `resolveProducerLink` is covered for a document that carries the reference (no DB read), a legacy CycloneDX row (link from `meta`), an SPDX row (no link), and an unreadable row (no link, enrichment proceeds).
- Two existing assertions that pinned the *old* contract — the served processed document carrying the producer's serial — were updated to assert the new one (`test/unit/bomService.test.ts`, `test/integration/bom-crud.test.ts`). Duplicate detection and the version-conflict refusal are unchanged and still green in `bomVersionConflict.test.ts`.
- Full unit suite green: 210 tests, plus `test/integration/bom-crud.test.ts` 4/4. `tsc --noEmit` clean. `test/integration/bom-diff.test.ts` fails 4/4 identically on `main` and on this branch in this sandbox: `spawn cyclonedx-cli ENOENT`, the binary is not installed here. It asserts nothing about serial numbers and this change does not touch `bomDiffService`.

Not covered by an automated test: a real two-run enrichment against BEAR, and rearm-core's in-place version bump landing on the same artifact. Both were verified by reading the code paths — `validateCycloneDxUpdate` compares the *uploaded* file's serial against `internalBom.id`, and the processed body keeps the producer's `version`.

## Breaking Changes

One accepted behaviour change: re-uploading an augmented download as an update of the same artifact now creates a **new** artifact, because its serial no longer matches. That flow was never supported — the augmented copy is not the producer's document.

**SPDX rows: the row identity now belongs to no stored document.** `meta.serialNumber` on an SPDX row is the CycloneDX identity rebom assigns it; the converted document it stores carries its own minted serial, and the uploaded document is SPDX. This is deliberate and no lookup is affected — every lookup keys on `meta.serialNumber`, which does not move. Recorded here so nobody later "fixes" it by pinning the converted document back to the row's serial, which would put two different documents under one identity again.

## Deployment / Operations

No migration and no backfill. Existing rows keep a processed document whose serial equals the producer's and have `processedSerialNumber` absent; readers must not assume the two differ. Those rows converge on their next enrichment run, which resolves the producer's identity from the row when the document does not carry the reference yet.

## Security

None directly. It removes a soundness problem in the other direction: a signature or BOM-Link check against the producer's identity no longer resolves to a document rebom augmented.

## Checklist

- [x] I have read and agree to the [Licensing of contributions](../CONTRIBUTING.md#licensing-of-contributions) terms: my contribution is licensed under the repository's license, and I additionally grant Reliza the right to relicense it (including for ReARM Pro)
- [x] Title follows conventional commits (feat, fix, chore, docs, refactor, build, ci)
- [x] Description clearly explains the change and context
- [x] Tests added/updated or rationale provided
- [ ] Documentation updated — no user-facing doc describes the storage identity model
- [x] No sensitive data committed (keys, tokens, secrets)
- [x] Backward compatibility considered

## Notes for Reviewers

- **D4 scope.** The producer reference is written for every processed document including specVersion < 1.4, where BOM-Link is not yet a defined concept. Confirmed schema-valid there by test. Kept unconditional by decision.
- **Pre-existing, found while writing that test, not fixed here:** `createRebomToolObject` always emits `type: "application"`, and for specVersion < 1.5 that object goes into the legacy `metadata.tools[]` array, whose schema forbids `type`. So the *stored processed copy* of any 1.2–1.4 upload already fails strict CycloneDX validation on `main` — ingest doesn't catch it because `validateBom` runs before augmentation. Measured: plain 1.3 valid, 1.3 + this PR's reference valid, 1.3 and 1.4 + the rebom tool **invalid**, 1.5 + the tool valid. Worth its own PR.
- **Raw substitute.** When the raw copy is gone and `rawBomMissing` makes us serve the processed document instead, that document now has a different serial from the row. Correct — it is a substitute, not the producer's bytes — and the `fetchProcessedSubstitute` docblock now says so.
- **SPDX rows** now have a `meta.serialNumber` that no stored document carries. That mirrors the CycloneDX case and no lookup is affected, but it is the one place where the row identity belongs to no artifact at all.

---


taleodor-claude:
## Live e2e on the scully sandbox

`rebom-backend` hot-swapped to an image built from this branch (`rebom-backend:local-serial1`, `505c2dca`), driven through the real ReARM path — `addArtifactManual` multipart upload, artifact `download`/`rawdownload`, BEAR enrichment with the sandbox's real integration. **28/28 checks pass.** The image build itself runs the full suite against a real Postgres with `cyclonedx-cli` present: **245/245**, including the four `bom-diff` integration tests that can't run in my shell.

### Ingest (10/10)

Producer serial `urn:uuid:8e1ff038…`:

| | |
|---|---|
| artifact `internalBom.id` | `8e1ff038…` — the producer's, unchanged |
| `meta.serialNumber` | `urn:uuid:8e1ff038…` |
| `meta.processedSerialNumber` | `urn:uuid:5ef7f4e8…` (distinct) |
| processed download `serialNumber` | `urn:uuid:5ef7f4e8…`, `version: 1` |
| processed download ref | exactly one `{type: bom, url: urn:cdx:8e1ff038-…/1, comment: "Source document as uploaded by the producer"}` |
| raw download | `urn:uuid:8e1ff038…`, no back-reference |

### Enrichment (12/12)

Enrichment fired on upload, then a forced second run. Four documents, four identities, one row:

```
producer   urn:uuid:8e1ff038-ef30-41b0-8423-812a2b2f448d   (raw, -raw tag)
upload     urn:uuid:5ef7f4e8-8d5d-4985-9d6e-32f3ae69370f   (enrichments[0], bare uuid tag)
run 1      urn:uuid:89ba252f-8dae-49a3-a7dc-33b68d1b2afb   (enrichments[1], -e1)
run 2      urn:uuid:4ac843f3-95e0-4dc8-b083-d814847ecee9   (enrichments[2], -e2)
```

`meta.processedSerialNumber` followed the pointer to `-e2` each time, `meta.serialNumber` never moved, and the twice-enriched document still carries **one** producer reference pointing at `urn:cdx:8e1ff038-…/1` — the link is refreshed, never chained to the previous run.

### rearm-core update semantics (6/6)

The compatibility claim I could previously only read:

- Uploading the same serial at `version: 3` to the same artifact keeps **the same artifact uuid**, bumps its version `2 → 3`, and leaves `internalBom.id` at the producer's serial. Three rebom rows under one identity, each with its own processed serial.
- Uploading the **augmented download** back (its serial is now the minted one) creates a **new** artifact and leaves the original untouched — the documented behaviour change, confirmed rather than assumed.

No migration was run and no existing row was rewritten; legacy rows converge on their next enrichment.

taleodor-claude:
## Review hardening applied, and re-verified live

`2ca5f008`. The mint now hangs off `processable`, not off augmentation — you're right that `processBomObj` has already sanitized, deduplicated and repaired dependencies before `augmentBomForStorage` runs, so gating on augmentation meant `AUGMENT_ON_STORAGE = false` would have shipped a deduplicated document under the producer's serial. `augmentBomForStorage` is back to pure augmentation, with a docblock saying why it must not mint. The verbatim non-processable path is untouched. One side benefit: forced re-enrichment no longer mints a serial it was about to discard.

Pinned by a test rather than by the constant: an upload with `augmentBomForStorage` stubbed to a passthrough still pushes a minted, back-linked document and records `processedSerialNumber`.

**Spec 1.3 validity — and a pre-existing bug it surfaced.** The reference is valid at 1.3, as expected. But the first version of that test failed, and not because of the reference:

| document at 1.3 / 1.4 | verdict |
|---|---|
| plain | valid |
| + this PR's minted serial and producer reference | **valid** |
| + `attachRebomToolToBom` (already on `main`) | **invalid** — `/metadata/tools/0 must NOT have additional properties` |
| the same tool at 1.5 | valid |

`createRebomToolObject` always emits `type: "application"`, and below 1.5 that object goes into the legacy `metadata.tools[]` array, whose `tool` schema forbids `type`. So the stored processed copy of any 1.2–1.4 upload already fails strict CycloneDX validation on `main` — ingest doesn't catch it because `validateBom` runs before augmentation. Out of scope here; the new test therefore asserts on a bare document so it pins the reference and not that bug. Happy to open a separate PR for it.

**SPDX row identity** now has its own paragraph in the Breaking Changes section, phrased so a later reader doesn't "fix" it by pinning the converted document back to the row's serial.

**Raw substitute** — no behaviour change, but `fetchProcessedSubstitute`'s docblock now records that a raw request answered this way returns a different serial from the row's, and that this is the honest answer rather than an oversight.

### Re-verified on the sandbox

Rebuilt (`rebom-backend:local-serial2`, in-build suite **247/247**), redeployed, and the whole live e2e re-run against the hardened image: **10/10 ingest, 12/12 enrichment, 6/6 rearm-core update semantics**. Four distinct identities on one row again (`33b0b4f2` producer → `4cb99df8` upload → `1aeac668` run 1 → `e04c2d92` run 2), the producer link still single and still pointing at the producer after two runs, and the same-serial `version: 3` upload still landing on the same artifact uuid.

Unit suite 210 green, `tsc --noEmit` clean.

PR #374, opened 2026-09-21
fix(oci): a manifest digest describes its content and nothing else

logicflakes:
`oras.PackManifest` was called with no `ManifestAnnotations`, so oras-go filled in `org.opencontainers.image.created` with `time.Now()` -- its own doc says to pass a fixed value for exactly this reason. Two pushes of byte-identical content therefore produced two different manifest digests, even though the layer blob is content-addressed and shared between them.

## Why it matters

That would be cosmetic if nobody kept the digest. They do: ReARM records the **manifest** digest against an artifact and resolves it later to serve the raw download. Our tags are content-addressed, so a second push of the same bytes re-points the tag at a new manifest and leaves the first untagged -- while a live row still addresses it by digest. An orphan that survives only because the registry runs with garbage collection off, and a 404 on an untouched artifact the day it does not.

## The change

Pinning the annotation to the epoch makes the digest a function of the content, so the second push is a genuine no-op and there is nothing to orphan. Nothing is lost by it: the annotation describes the manifest rather than the artifact, and callers keep their own creation time next to the digest they store.

The pack step is split into `packArtifactManifest` so the property is testable without a registry. The call site comes out a line shorter.

## On the test, because the obvious one does not work

The timestamp has **one second** of resolution, so two packs inside a test land in the same second and agree even with the fix reverted. The first version of this test **passed five times out of five against the bug it existed to catch.**

So the guard asserts the annotation that actually went into the manifest, and the digest-equality test sleeps a real second -- load-bearing rather than a flake workaround, and skipped under `-short`. Reverted, both fail three times out of three.

## Verified live

Against the sandbox registry, not only in unit tests -- same bytes pushed twice, more than a second apart:

| | push #1 | push #2 | |
|---|---|---|---|
| before, small | `2091ddad` | `db233b02` | differ |
| before, large (zstd) | `9f03ca60` | `25d45c78` | differ |
| after, small | `8d6d6ac8` | `8d6d6ac8` | **identical** |
| after, large (zstd) | `8d2e74b0` | `8d2e74b0` | **identical** |

Both the plain and the compressed path, and distinct content still packs to distinct manifests. The "before" column is real rather than reconstructed: an earlier build went in from the wrong tree, which turned out to be a free control.

`go vet`, `go test`, `gofmt` clean. The import reorder in `oras.go` is gofmt -- the file was already unformatted on main and editing it made gofmt notice.

## Deliberately not in scope

**Forward only.** Manifests already pushed with a wall-clock timestamp keep their digests, so a row pointing at one can still be orphaned by a later push of the same bytes. Covering those means not re-pushing when the content is already there -- and `PushArtifact` is shared with callers whose tags are **not** content-addressed, where skipping on a tag hit would silently return a stale manifest for new content. The safe form compares the existing manifest's layer digest against what is about to be pushed. That is a change to push behaviour and wants its own review.

This should land before anyone enables registry garbage collection. Enabling GC needs its own design pass regardless: a monthly repository is shared across organizations, so a content-addressed blob may only be reclaimed when no artifact in any organization still carries its content address.


logicflakes:
Layer 1 review run on the branch (`8909f5b2`). Two real findings, both about the tests rather than the fix.

## The guard did not cover the call site

Every test called `packArtifactManifest` directly, so reverting `PushArtifact` to an inline `oras.PackManifest` -- **exactly the bug** -- left the whole suite green. Verified before fixing, and again after.

`TestPushArtifactSendsAFixedCreatedAnnotation` now drives the real `PushArtifact` against a minimal in-process OCI registry and reads the `created` annotation off the manifest that was actually pushed. The fake registry is the smallest surface `oras.Copy` needs; the digest header is computed inline rather than through `go-digest`, so `go.mod` stays untouched.

Worth saying plainly: this is the **second** time on this change that a test passed against the bug it existed to catch. The first was comparing two digests packed inside the same wall-clock second. Both were caught by reverting the fix and re-running rather than by reading the test.

## The tests never ran

`build-oci-service` only invokes the image build, and the Dockerfile ran `go build` with no `go test` -- so every test in this package, including the pre-existing `validation_test.go`, executed only when a human typed `go test`. My comment claiming a guard "fails the moment the annotation stops being fixed" was true locally and false in CI.

The build now runs them. Verified by building the image: `go test ./...` reports `ok` inside the build stage.

## Smaller items

- Dropped `TestCreatedAnnotationIsPinnedToTheEpoch` -- it compared a `const` against its own literal, which the compiler folds. It could only fail if someone edited the constant, which is a change detector rather than a test. The real protection is that a non-RFC-3339 value makes oras-go refuse every push.
- Corrected the premise in the comment on `manifestCreatedAnnotationValue`: it said our tags are content-addressed, true of retained raw uploads (`rearm-raw-<hash>`) but not of downloadable artifacts, which are tagged with the artifact UUID and re-pushed under that tag on re-upload. The fix applies to both; the sentence now says so.

## Checked and clean

`packArtifactManifest` narrowing to `content.Pusher` loses nothing -- it is `PackManifest`'s own parameter type, and the `Exists()` short-circuit is recovered by a dynamic type assertion inside oras-go. The `Errorw` at the call site is retained verbatim. And **nothing in this repo reads `org.opencontainers.image.created`**: no retention, no listing, nothing sorting by manifest age, and zot ships with `gc` and `dedupe` off and no retention block. That was the consequence I was most worried about, since no test covers it.

## One residual, for the record

Determinism is bounded by the blob, which no test pins. The manifest digest still covers the zstd-compressed bytes and the `io.reliza.*` compression annotations, and the compress/skip decision is media-type driven. A zstd encoder change or an edit to `compressibleMimeTypes` would make the same original content pack to a new manifest digest under the same tag -- the orphan scenario returning through a different door. Nothing to do now; noting it so the guarantee is not over-read.

PR #376, opened 2026-09-21
feat(release): declutter the release page, and make the export modal say what it exports

logicflakes:
## Part A -- the release page

FDA-Readiness-1 landed three panels at the top of the Components tab. The operator
walkthrough found the first two were read as **one** control often enough that the
explainer had to open by saying they were not, and all three sat between the
component table and the artifact list where an operator preparing a submission had
to already know they were there.

- **The read-only device support window is removed, not relabelled.** It is declared
  on the product component and resolved per shipment on Distribution -- both surfaces
  already show it -- and a third read-only copy on a page that cannot edit it bought
  nothing but the confusion. `loadInheritedDeviceWindow` and its per-release round
  trip go with it.
- **Release lifecycle dates move to Meta**, under Notes, for PRODUCT and COMPONENT
  releases alike. The explainer keeps the "separate from the device support window"
  sentence, reworded: it no longer says "above", and now says where the window is.
- **New PRODUCT-gated Support tab** between VEX and Meta: the assessment
  justification, and the release-scoped coverage gauge with a link to the rows
  behind it. Components is the inventory of what this release *is*; this tab is what
  we *claim* about it.
- The gauge **stays duplicated** above the SBOM list -- that copy is the number that
  moves as the operator attests. Both render `sbomCoverageDisplay`, so they cannot
  disagree.
- **Attestation editing stays in BOM Components > SBOM.** A control that writes a
  regulatory claim across thousands of components belongs next to the rows it
  writes.

The Components tab is now **byte-identical to `origin/main`'s** -- asserted by
construction while making the change, not by eye.

## Part B -- the export modal

"Support" is a BOM **type** now, after CLE, not three formats hidden among the SBOM
encodings. Picking it swaps the format list and drops the BOM-shaping controls,
which is **structural**: the five `v-if="!isFdaDocumentExport"` gates are gone, and
so is the predicate. That mechanism is what once let the artifact-coverage filter
stay toggleable while the addendum ignored it.

Two new toggles beside "Top Level Dependencies Only": **Include support metadata**
(defaults to the org setting; disabled *with a reason* when the org has it off) and
**Include internal metadata** (defaults off). Sent via a document with the two
arguments, falling back to the flagless one on a schema-drift rejection -- a backend
that does not declare them rejects during *validation*, before any resolver runs, so
nothing was exported and the retry is not a second write.

## Part D -- labels

"FDA support addendum" -> "Support addendum"; "FDA submission and labeling text" ->
"Submission and labeling text" with the purpose moved into a tooltip rather than
deleted. Identifiers untouched; the FDA citations inside generated documents stay --
they are citations, for a reader who wants them.

## Review found a blocker the whole green suite missed

`watch(exportBomType, ...)` was declared **85 lines above** `const exportBomType`.
`watch()` evaluates a bare-ref source eagerly, so setup threw
`Cannot access 'exportBomType' before initialization` and the **entire release page
rendered nothing** -- every tab. `vite build`, 748 vitest assertions and eslint were
all green, because nothing in the suite mounts the component.

So the fix is two things. The watch moved below its ref, **and** the rules it drives
moved out into `utils/exportFormatSelection.ts` and `utils/exportMetadataFallback.ts`
with specs that *call* them. The wiring specs keep only what is genuinely about the
component -- that it delegates rather than holding a second copy -- plus a pin that
the watch is declared below its ref.

Other findings, all fixed:

- The drift latch was set on the **rejection**, not on the retry succeeding. The same
  mutation carries two enum arguments that drift identically, so one unrelated bad
  request could have disabled the options for the session -- and every later export
  then went out flagless with **no** warning while both switches still read as live.
- "Attest components in BOM Components" routed to the **HBOM** sub-tab on a hardware
  product release, which has no gauge, no filter and no attest control at all.
- Adding `v-model` to the outer `n-tabs` created two failures reachable with one click
  on the prev/next chevrons: a stranded `activeTab` with no matching pane (blank tab
  strip), and lazily-loaded tabs sitting empty because nothing fires `update:value` on
  a programmatic change.
- `inProductsLoaded` / `hbomLoaded` were never reset on release change (pre-existing),
  so the tab reload above was calling loaders that early-returned.
- Every sliced assertion now proves its slice is **non-empty** first: three specs had
  runs of `not.toContain` over a slice an empty string satisfies in silence.

## Validation

- `npm run test` **763 passed / 0 failed**. (Baseline on a clean tree is 747, not the
  858 quoted in the brief -- 11 spec *files* fail to collect because every co-located
  Pro-schema loader reads only `schema.graphqls` while `MatchOperator` now lives in
  `programmatic.graphqls`. Pre-existing, reproduced with these changes stashed, filed
  as `t20260921-120438-20991`. It also silently disables `validate-graphql.mjs`'s Pro
  checks, so both export documents were hand-validated against the concatenated
  three-file schema: FULL and CORE valid on Pro, FULL correctly rejected by CE.)
- `vite build` clean.
- **Browser-probed live** on the claude2 sandbox (UI `...flags.3`, backend `...flags.1`):

| release | Support tab | Components clean | Meta has dates | Support: justification / gauge / link |
|---|---|---|---|---|
| product | yes | yes | yes (2 pickers) | y / y / y |
| component | **no** | yes | yes (2 pickers) | n/a |

- **The export modal, driven by clicking it** (`export-modal-probe.mjs`):

```
BOM types offered: SBOM / OBOM / VDR / VEX / CLE / Support
Support formats:   Support addendum (CSV) / Support addendum (PDF) / Device support statement (PDF)
  Support addendum (CSV)         -> downloaded "fda-support-addendum-9000.1.0.csv"
  Support addendum (PDF)         -> downloaded "fda-support-addendum-9000.1.0.pdf"
  Device support statement (PDF) -> downloaded "device-support-statement-9000.1.0.pdf"
format after Support -> SBOM: CycloneDX 1.6 (JSON)   (the reset works; not blank)
has "Include support metadata": true
has "Include internal metadata": true
org-disabled hint shown: true
CSV shows the inert-options note: true
page errors: none
```

On a COMPONENT release the Support type offers the two addenda only -- the device
statement is withheld. All three documents download for real; file names keep the
`fda-` prefix on purpose, since Part D renamed labels rather than identifiers.

Running this needed a sandbox fix first: `rearm-oci-artifact` had been
CrashLoopBackOff for 15 days, so every BOM egress returned 500. Fixed permanently
through the normal pipeline; unrelated to this change.

## Note for the reviewer

This targets `fda-readiness-1-dev`, which is **22 commits behind `main`** and was
deliberately *not* merged here -- that reconciliation is 24 conflicted files, 23 of
them backend CE-mirror adjudication between two independent syncs of the same
feature, and belongs in its own PR (`t20260921-114344-17632`). Only one conflict is
in the UI.

The two export arguments reach CE only at its next sync, which is why the modal
carries the drift fallback.

Backend PR: relizaio/rearm-saas#567


logicflakes:
## Now validated live end to end

The export half of the probe could not run earlier (`rearm-oci-artifact` CrashLoopBackOff for
15 days). **Fixed** -- it was pointed at a registry that does not hold the blobs, with a Secret
that does not exist. Every BOM egress works again, so the modal has been driven by clicking it.

**The export modal, on a PRODUCT release:**

```
BOM types offered: SBOM / OBOM / VDR / VEX / CLE / Support
Support formats:   Support addendum (CSV) / Support addendum (PDF) / Device support statement (PDF)
  Support addendum (CSV)         -> downloaded "fda-support-addendum-9000.1.0.csv"
  Support addendum (PDF)         -> downloaded "fda-support-addendum-9000.1.0.pdf"
  Device support statement (PDF) -> downloaded "device-support-statement-9000.1.0.pdf"
format after Support -> SBOM: CycloneDX 1.6 (JSON)   (the reset works; not blank)
has "Include support metadata": true
has "Include internal metadata": true
org-disabled hint shown: true
CSV shows the inert-options note: true
page errors: none
```

**On a COMPONENT release:** Support formats are the two addenda only -- the device statement is
withheld, as designed.

All three documents download for real. File names still carry `fda-` on purpose: Part D renamed
labels, not identifiers.

**Release page**, re-confirmed on the final build:

| release | Support tab | Components clean | Meta has dates |
|---|---|---|---|
| product | yes | yes | yes (2 pickers) |
| component | **no** | yes | yes (2 pickers) |

New probe: `ai-agents/probes/fda-readiness-1/export-modal-probe.mjs` in the backend repo (it
drives this UI). Walkthrough: `ai-plans/release-declutter-export-flags-walkthrough.md`.

Two selector notes are written into the probe because each cost a run: naive-ui puts `name` on
the inner input rather than the radio-group wrapper, and Playwright's `has-text` is a substring
match -- `"Support"` matches `"Support addendum (CSV)"` -- so radios are indexed by exact
trimmed label.

logicflakes:
## Two-layer review on the delta (`848a2892`), and what it found

Layers 1 and 2 had already run on this branch up to `604ede71`. `848a2892` (support switch absent when the org setting is off, both switches default off) landed after that, so it went through both layers on its own. Fixes are in `48bafbac`.

No behaviour change to what was decided: the switch is still absent when the organization does not publish attestations, still defaults off, and still sends `null` rather than `false` when it was never shown. What changed is that all three can now fail a test, and the copy around them stopped describing a modal that no longer exists.

### Both layers, independently: the wire rule was untested

The three-outcome rule lived as a ternary inside a GraphQL variables literal, and **no spec referenced `includeSupportMetadata` or `orgSupportInjectionEnabled` at all.** Dropping the ternary back to `includeSupportMetadata.value` re-ships the exact `false`-instead-of-`null` bug the commit message argues at length must not happen, with all 763 tests green.

It is now `supportMetadataArg(offered, chosen)` in `utils/exportMetadataFallback` with a three-case spec -- this repo's documented answer to a rule nothing can mount, and the same reason `exportFormatSelection.ts` exists on this branch. The third case gets its own test: a stale switch left ON by an org that then turned the setting off must still send `null`, not `true`.

Five wiring assertions tie the two statements of the gate together. It is stated twice by necessity -- once to render the switch, once to shape the argument -- and nothing in the framework keeps them agreeing.

### Copy that counted the switches

The inert-format note read "**These two options** do not change a CSV or EXCEL export" -- shown under a **single** switch on every organization with the setting off, which is the default. Now "the metadata options above". Three more statements of the old modal went with it: an HTML comment about "two live-looking switches", a computed's jsdoc still arguing this form can request a disclosure the organization disabled (it cannot -- that validation is now an API-only path, which the jsdoc says instead), and the docs page.

### The docs contradicted the new default in their normative section

"Every JSON BOM ReARM serves carries `reliza:support:disclosure` ... with one of two values. **Read it before drawing a conclusion from an absent property**" -- false for the most common export there is after this change: org publishes attestations, operator clicks Export without touching the switch, document comes back with no marker at all. The table has a third row now, and the section says the marker is about **who decided**.

### The CE sync-lag hint came back

"Absent, not disabled" was the ruling about an organization that has the setting **off** -- the settings screen is where that is decided and it is reachable. A backend that does not **declare** the setting is a different case with nothing the operator can go and change, and it lost its explanation by accident. It has one again, gated on `orgSupportInjectionSupported` alone.

### Known, not fixed here

The gate reads the **currently selected** org's settings while the backend decides with the **release's** org. I could find no route that puts this page on a foreign-org release, and the sibling export paths in this file have the same shape, so aligning them is its own change.

### Validation

- `npm run test` **771 passed** (763 + 8), 0 failed. The 11 spec *files* that fail to collect are the pre-existing dark schema-drift loaders (`t20260921-120438-20991`), unchanged by this branch.
- `vite build` clean. eslint: the same 10 pre-existing indent errors, shifted by the lines added above them -- no new ones.
- Deployed to the claude2 sandbox as `2026-09-release-page-declutter-export-flags.5` and **browser-probed**, because a green build proves nothing about mounting in this repo:
  - release page renders, all ten tabs including Support; no page errors
  - support switch offered: `false`, correct for this DISABLED org, and the org read now *prints the value it read* rather than defaulting to `false` on failure
  - the note does not count the switches: `true`
  - CE sync-lag hint correctly absent on a server that declares the setting: `true`
  - all three support documents still download; device statement still withheld on a COMPONENT release

### CI

The docs-site job on `48bafbac` failed on an infrastructure flake, not on this change: `error fetching GitHub OIDC token ... i/o timeout`, after which cosign fell back to a device flow and hit `expired_token`. **`Build And Push ReARM UI` and `Build And Push ReARM Backend` both succeeded** and the images deployed. I could not re-run the failed job -- `Resource not accessible by personal access token`. Worth a re-run before merge.

